PR #3, opened 2016-12-27
Add support for verifying the SSL certificate presented by the backing mongo

tolsen:


tolsen:
cc @erh

tolsen:
good catch. Should I have those binaries continue to skip SSL verification or should I add an option to both of them for specifying a CA file?

erh:
I think just having them continue to skip ssl verification is easiest and fine for now

tolsen:
Ok. done

PR #7, opened 2017-02-09
Handle error response without a message

tolsen:
@erh 

It is possible for InterceptClientToMongo to return a non-nil error with a nil message.  This PR handles that case.

### DIFF
--- a/proxy.go
+++ b/proxy.go
@@ -162,7 +162,13 @@ func (ps *ProxySession) doLoop(pooledConn *PooledConnection) (*PooledConnection,
 
 		m, respInter, err = ps.interceptor.InterceptClientToMongo(m)
 		if err != nil {
-			if m == nil || !m.HasResponse() {
+			if m == nil {
+				if pooledConn != nil {
+					pooledConn.Close()
+				}
+				return nil, err
+			}
+			if !m.HasResponse() {
 				// we can't respond, so we just fail
 				return pooledConn, err
 			}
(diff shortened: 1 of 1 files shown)


erh:
In the m == nil case we have to close the connection and return a nil connection.
Otherwise we're pinning the socket.

tolsen:
By pinning do you mean that we would be reusing the same pooled connection in the for loop here: https://github.com/erh/mongonet/blob/master/proxy.go#L271-L282 ?

The way I've written it now the connection will be closed here: https://github.com/erh/mongonet/blob/master/proxy.go#L275 .   And then ProxySession.Run() will return.  I assumed that was the desired behavior when InterceptClientToMongo returns an error regardless of whether m == nil.  Is that wrong?

erh:
I see, I guess it won't get pinned but I think we should still close it right when we get the error.
If we know we think its bad, we should close.

erh:
We don't want to change the behavior in the other case
I would leave what was there originally and add a
```
if m == nil {
  if pooledConnection != nil { 
     pooledConnection.Close()
  }
  return nil, err
}
```

tolsen:
Should I also mark the connection as bad?  Otherwise we're just sending it back to the pool to be reused.

erh:
My current thinking is no.
If its an io error to the user then its bad.
If its an io error to the backend it'll get handled elsewhere.

PR #10, opened 2017-03-03
Return errmsg instead of $err for OP_QUERY and OP_GET_MORE

tolsen:


tolsen:
@erh Just pinging you in case you missed this.

PR #12, opened 2017-03-10
Reduce TLS handshake error to WARN level

tolsen:
It's been requested that we reduce TLS handshake errors to a warning.  Some users are trying to connect without SSL and the errors are adding noise to our error rate reporting.


tolsen:
ping @erh.  When you get a chance.

PR #17, opened 2017-05-26
Return error in Read/Write on TLS connection corruption

louisaberger:


louisaberger:
Discussed with @tolsen already, merging directly.

PR #18, opened 2017-06-08
Index out of range panic in namespace.go

louisaberger:
https://jira.mongodb.org/browse/CLOUDP-22737


tolsen:
We should probably keep links to private Jira tickets out of pull requests on a public repository (same for branch names)

louisaberger:
Good point! Going to remove this PR and change those.

PR #21, opened 2017-12-15
OP_MSG support

tolsen:


wire.go
```diff
@@ -132,3 +133,12 @@ type CommandReplyMessage struct {
 	Metadata     SimpleBSON
 	OutputDocs   []SimpleBSON
 }
+
+// OP_MSG
+// Note that checksum is not implemented
```
louisaberger: What is this?

wire_message.go
```diff
@@ -0,0 +1,185 @@
...
+
+	(*loc) += int(bs.Body.Size)
+
+	return bs, nil
+}
+
+func parseDocumentSequenceSection(buf []byte, loc *int) (MessageMessageSection, error) {
+	dss := &DocumentSequenceSection{}
+
+	expectedSizeRemaining := readInt32(buf[*loc:])
```
louisaberger: Do we need to check here that len(buf) is >= *loc+4 so that readInt32 doesn't panic?

wire_message.go
```diff
@@ -0,0 +1,185 @@
...
+	docs := make([]SimpleBSON, 0)
+	docNum := 0
+	for expectedSizeRemaining > 0 {
+		doc, err := parseSimpleBSON(buf[*loc:])
+		if err != nil {
+			return dss, err
+		}
+
+		if expectedSizeRemaining < doc.Size {
+			return dss, NewStackErrorf("invalid Document Sequence section -- size of document %v (starting from 0) overruns end of section", docNum)
```
louisaberger: Should we log the `expectedSizeRemaining` and `doc.Size` here too?

wire.go
tolsen: The checksum is an optional part of OP_MSG.  It would be at the end of the struct if it were implemented

tolsen:
@louisaberger comments addressed!

tolsen:
@erh I'm going to merge but please feel free to take a look and if you have any feedback we can address it in another PR!

PR #23, opened 2017-12-19
Refactor out server

erh:
This is not 100% complete, but its a good staging point and don't want to keep diverging from the main line.

### DIFF
--- a/wire_message.go
+++ b/wire_message.go
@@ -138,13 +138,13 @@ func parseBodySection(buf []byte, loc *int) (MessageMessageSection, error) {
 func parseDocumentSequenceSection(buf []byte, loc *int) (MessageMessageSection, error) {
 	dss := &DocumentSequenceSection{}
 
-	expectedSizeRemaining := readInt32(buf[*loc:])
-
-	if expectedSizeRemaining < 4 {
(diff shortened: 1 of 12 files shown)


wire_message.go
```diff
@@ -138,13 +138,13 @@ func parseBodySection(buf []byte, loc *int) (MessageMessageSection, error) {
 func parseDocumentSequenceSection(buf []byte, loc *int) (MessageMessageSection, error) {
 	dss := &DocumentSequenceSection{}
 
-	if len(buf[*loc:]) < 4 {
+	expectedSizeRemaining := readInt32(buf[*loc:])
```
tolsen: This file seems to be missing changes I made to my op-msg branch after you branched from it.  Possibly a bad merge from master?

erh: Is that the only one or is there a diff or are there more that are missing?
I thought I had everything...

erh:
I pulled in wire_message.go from master.
Anything else look off?

tolsen:
ugghh.. this probably happened because I deleted the op-msg branch after merging into master.  Sorry.  It looks like we have all the changes now.

PR #28, opened 2019-01-25
Set pooled connection to be bad if there is an error during sending

tolsen:


proxy.go
```diff
@@ -244,6 +244,7 @@ func (ps *ProxySession) doLoop(pooledConn *PooledConnection) (*PooledConnection,
 
 	err = SendMessage(m, mongoConn)
 	if err != nil {
+		pooledConn.bad = true
```
louisaberger: Should we have the same on line 277?

tolsen: No.  Line 277 is sending a message to the client (the end user) so that is not part of the connection pool (which is to the MTM)

PR #32, opened 2019-07-01
Dynamic certificates loading

TomerYakir:
Manually tested that when passing additional certificates, the correct certificate is shown according to SSLServerName.
Also tested backwards compatibility with existing application

### DIFF
--- a/server.go
+++ b/server.go
@@ -29,8 +29,6 @@ func (s *SyncTlsConfig) getTlsConfig() *tls.Config {
 }
 
 func (s *SyncTlsConfig) setTlsConfig(sslKeys []*SSLPair, cipherSuites []uint16, minTlsVersion uint16, fallbackKeys []SSLPair) error {
-	s.lock.Lock()
-	defer s.lock.Unlock()
 	certs := []tls.Certificate{}
 	for _, pair := range fallbackKeys {
 		cer, err := tls.LoadX509KeyPair(pair.Cert, pair.Key)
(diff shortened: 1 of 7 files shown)


server.go
```diff
@@ -10,14 +10,67 @@ import "time"
...
+}
+
+func (s *SyncTlsConfig) getTlsConfig() *tls.Config {
+	s.lock.RLock()
+	defer s.lock.RUnlock()
+	return s.tlsConfig
+}
+
+func (s *SyncTlsConfig) setTlsConfig(sslKeys []*SSLPair, cipherSuites []uint16, minTlsVersion uint16, fallbackKeys []SSLPair) error {
+	s.lock.Lock()
```
tolsen: what do you think about moving the grabbing of the lock to just before we set s.tlsConfig?

server.go
```diff
@@ -10,14 +10,67 @@ import "time"
...
 type ServerConfig struct {
 	BindHost string
 	BindPort int
 
 	UseSSL        bool
 	SSLKeys       []SSLPair
-	MinTlsVersion uint16 // see tls.Version* constants
+	SyncTlsConfig *SyncTlsConfig
 
+	MinTlsVersion      uint16
```
tolsen: lost a comment here?

server.go
TomerYakir: good idea - fixed

server.go
TomerYakir: readded

TomerYakir:
This is partially backwards compatible through the "fallback keys" mechanism.
The existing Atlas/MMS code won't work as-is and would require some minor modifications to work, specifically calling `proxy.InitializeServer()` and `proxy.OnSSLConfig()` before invoking `proxy.Run()`.

PR #36, opened 2020-02-16
Use go driver bson lib

TomerYakir:


server_test.go
```diff
@@ -177,35 +237,44 @@ func TestServer(t *testing.T) {
...
+	client, err := mongo.NewClient(opts)
 	if err != nil {
-		t.Errorf("cannot dial: %s", err)
-		return
+		t.Errorf("cannot create a mongo client. err: %v", err)
 	}
-	defer session.Close()
 
-	coll := session.DB("test").C("bar")
+	ctx := context.Background()
```
tolsen: maybe we should have a timeout for the test?

TomerYakir: added!

server_test.go
```diff
@@ -427,13 +428,15 @@ func TestServerWorkerWithContext(t *testing.T) {
 			t.Errorf("cannot create a mongo client. err: %v", err)
 			return
 		}
-
-		ctx := context.Background()
+		
+		ctx, cancelFunc := context.WithTimeout(context.Background(), 5 * time.Second)
```
tolsen: Is this overwriting the prior cancelFunc?  If so, will the previous defer run properly?  When is the name looked up?

TomerYakir: changed

PR #37, opened 2020-02-26
support unacked writes

TomerYakir:


proxy.go
```diff
@@ -205,10 +205,11 @@ func (ps *ProxySession) doLoop(pooledConn *PooledConnection) (*PooledConnection,
 	}
 
 	var respInter ResponseInterceptor
+	expectingResponse := true
 	if ps.interceptor != nil {
 		ps.interceptor.TrackRequest(m.Header())
 
-		m, respInter, err = ps.interceptor.InterceptClientToMongo(m)
+		m, respInter, expectingResponse, err = ps.interceptor.InterceptClientToMongo(m)
```
tolsen: does the if on line 220 need to be modified too?

TomerYakir: I don't think so - that code deals with the pre-send to mongo part of the flow which shouldn't be changed

PR #39, opened 2020-02-26
Support unacked writes (mgo)

TomerYakir:


proxy.go
```diff
@@ -204,10 +204,11 @@ func (ps *ProxySession) doLoop(pooledConn *PooledConnection) (*PooledConnection,
 	}
 
 	var respInter ResponseInterceptor
+	expectingResponse := true
 	if ps.interceptor != nil {
 		ps.interceptor.TrackRequest(m.Header())
 
-		m, respInter, err = ps.interceptor.InterceptClientToMongo(m)
+		m, respInter, expectingResponse, err = ps.interceptor.InterceptClientToMongo(m)
```
tolsen: same comment as other PR.  does the if on line 219 need to modified too?

session.go
```diff
@@ -31,6 +33,10 @@ func (s *Session) Logf(level slogger.Level, messageFmt string, args ...interface
 	return s.logger.Logf(level, messageFmt, args...)
 }
 
+func (s *Session) SetRemoteAddr(v net.Addr) {
```
tolsen: why do we need to add this?

proxy.go
TomerYakir: https://github.com/erh/mongonet/pull/37#discussion_r384983311

session.go
TomerYakir: this was part of that commit - https://github.com/erh/mongonet/pull/39/commits/73ab45ff832fa959e6b6873ceef7c2961f2f944d. I'm not entirely sure why it's showing in this PR.

tolsen: This is a red flag as to something being wrong with the branch.  Could you please investigate?

TomerYakir:
will investigate what's wrong with the branch and open a new PR

PR #53, opened 2020-09-27
Circular BSON Error fix

TomerYakir:


louisaberger (approved):

divjotarora:
Sorry to jump in here, but would it make sense to add a test in `bsonutil_test.go` for the case where we're fixing `a.x` in the document `{x: 1, a: []}` and the additional case you mentioned on Slack for walking down a nested `bson.D` in an array (e.g. fixing `a[].b.c`)?

TomerYakir:
> Sorry to jump in here, but would it make sense to add a test in `bsonutil_test.go` for the case where we're fixing `a.x` in the document `{x: 1, a: []}` and the additional case you mentioned on Slack for walking down a nested `bson.D` in an array (e.g. fixing `a[].b.c`)?

I added that test case to the proxy code base but it does make sense to add it to the mongonet code base as well.

TomerYakir:
@divjotarora added the test cases and confirmed both pass with the fix and that both fail without!

PR #55, opened 2020-10-01
Go driver connection pool

TomerYakir:


divjotarora:
test

divjotarora (commented):
test

proxy.go
```diff
@@ -187,127 +209,195 @@ func (ps *ProxySession) respondWithError(clientMessage Message, err error) error
...
 	}
 
 	if !m.HasResponse() {
-		return pooledConn, nil
+		return mongoConn, nil
 	}
-
-	defer pooledConn.Close()
+	defer mongoConn.Close()
+	// TODO - should probably return nil for mongoConn so we don't attempt to close twice
```
tolsen, line 302: is this TODO important to do before merging?

server_test.go
```diff
@@ -164,6 +164,17 @@ func (mss *MyServerSession) handleMessage(m mongonet.Message) (error, bool) {
 		if cmdName == "insert" {
 			return mss.handleInsert(mm, cmd, ns), false
 		}
+		if cmdName == "find" {
```
tolsen, line 167: changing all these ifs to a switch statement that switches on cmdName may compile to faster code

server_test.go
```diff
@@ -164,6 +164,17 @@ func (mss *MyServerSession) handleMessage(m mongonet.Message) (error, bool) {
...
+			ns := fmt.Sprintf("%s.%s", db, cmd[0].Value.(string)) // TODO: check type cast?
```
tolsen, line 168: should we address the TODO?

tolsen (changes requested):

iwysiu (approved):
lgtm with comments

proxy.go
```diff
@@ -187,127 +209,195 @@ func (ps *ProxySession) respondWithError(clientMessage Message, err error) error
...
 	var respInter ResponseInterceptor
 	if ps.interceptor != nil {
 		ps.interceptor.TrackRequest(m.Header())
-
 		m, respInter, err = ps.interceptor.InterceptClientToMongo(m)
 		if err != nil {
 			if m == nil {
-				if pooledConn != nil {
-					pooledConn.Close()
+				if mongoConn != nil {
```
divjotarora, line 262: Question: why are we calling `mongoConn.Close()` in just this error case? Other error cases do `return mongoConn, err` and I think the `DoLoopTemp` function calls `mongoConn.Close()` if the error is non-nil. Can we do the same here?

proxy.go
```diff
@@ -187,127 +209,195 @@ func (ps *ProxySession) respondWithError(clientMessage Message, err error) error
...
 			}
 			if !m.HasResponse() {
 				// we can't respond, so we just fail
-				return pooledConn, err
+				return mongoConn, err
 			}
 			err = ps.RespondWithError(m, err)
 			if err != nil {
-				return pooledConn, NewStackErrorf("couldn't send error response to client %v", err)
+				return mongoConn, NewStackErrorf("couldn't send error response to client %v", err)
```
divjotarora, line 273: Could we preserve the original error and include it in this message like this:
```
if respondErr := ps.RespondWithError(m, err); respondErr != nil {
    return mongoConn, NewStackErrorf("couldn't send error response to client; original error: %v, error sending response: %v, err, respondErr)
}
```

This way, it'll be included in the `WARN` verbosity log in `DoLoopTemp` instead of only being logged at `INFO` verbosity in `RespondWithError`.

proxy.go
```diff
@@ -187,127 +209,195 @@ func (ps *ProxySession) respondWithError(clientMessage Message, err error) error
...
+}
+
+func extractTopology(mc *mongo.Client) *topology.Topology {
+	e := reflect.ValueOf(mc).Elem()
+	d := e.FieldByName("deployment")
+	d = reflect.NewAt(d.Type(), unsafe.Pointer(d.UnsafeAddr())).Elem() // #nosec G103
+	return d.Interface().(*topology.Topology)
+}
+
+func (ps *ProxySession) getMongoConnection() (driver.Connection, error) {
```
divjotarora, line 236: Seems like this is only called in one location and the caller immediately wraps it in `MongoConnectionWrapper`. Can we change this to directly return `(*MongoConnectionWrapper, error)` instead?

proxy.go
```diff
@@ -187,127 +209,195 @@ func (ps *ProxySession) respondWithError(clientMessage Message, err error) error
...
+	topo := extractTopology(ps.proxy.MongoClient)
```
divjotarora, line 237: This could be a potentially expensive reflection-based operation. Would it be possible to store this directly on `ps.proxy` so it's only done once?

proxy.go
```diff
@@ -187,127 +209,195 @@ func (ps *ProxySession) respondWithError(clientMessage Message, err error) error
...
+			return mongoConn, NewStackErrorf("got error sending response to client from conn %v: %v", mongoConn.conn.ID(), err)
 		}
-
+		logTrace(ps.proxy.logger, ps.proxy.config.TraceConnPool, "sent back data to user from mongo conn %v", mongoConn.conn.ID())
 		if ps.interceptor != nil {
 			ps.interceptor.TrackResponse(resp.Header())
 		}
 
 		if !inExhaustMode {
-			return nil, nil
```
divjotarora, line 286: What was the motivation for changing these from `nil` to `mongoConn`?

server_test.go
```diff
@@ -404,12 +372,20 @@ func checkClient(opts *options.ClientOptions) error {
 	if err := client.Connect(ctx); err != nil {
 		return fmt.Errorf("cannot connect to server. err: %v", err)
 	}
-	client.Disconnect(ctx)
+	coll := client.Database("test").Collection("test")
+	docIn := bson.D{{"foo", 17}}
+	if _, err = coll.InsertOne(ctx, docIn); err != nil {
+		return err
+	}
+	err = client.Disconnect(ctx)
```
divjotarora, line 380: nit: This and the error handling below is equivalent to `return client.Disconnect(ctx)`.

proxy.go
```diff
@@ -187,127 +209,195 @@ func (ps *ProxySession) respondWithError(clientMessage Message, err error) error
...
+		logTrace(ps.proxy.logger, ps.proxy.config.TraceConnPool, "reading data from mongo conn %v", mongoConn.conn.ID())
+		ret, err := mongoConn.conn.ReadWireMessage(ps.proxy.Context, nil)
 		if err != nil {
-			pooledConn.bad = true
-			return nil, NewStackErrorf("got error reading response from mongo %v", err)
+			mongoConn.bad = true
+			return mongoConn, NewStackErrorf("error reading wire message from mongo conn %v: %v", mongoConn.conn.ID(), err)
+		}
+		logTrace(ps.proxy.logger, ps.proxy.config.TraceConnPool, "read data from mongo conn %v", mongoConn.conn.ID())
+		resp, err := ReadMessageFromBytes(ret)
```
divjotarora, line 315: Can you leverage the existing `ReadMessage` function by using [`bytes.NewReader`](https://golang.org/pkg/bytes/#NewReader)?

proxy.go
```diff
@@ -1,62 +1,87 @@
...
+func logTrace(logger *slogger.Logger, trace bool, format string, args ...interface{}) {
+	if trace {
+		logger.Logf(slogger.DEBUG, format, args...)
+	}
 }
 
-func NewMongoError(err error, code int, codeName string) MongoError {
-	return MongoError{err, code, codeName}
+// MongoConnectionWrapper is used to wrap the driver connection so we can explicitly expire (close out) connections on certain occasions that aren't being picked up by the driver
+type MongoConnectionWrapper struct {
```
divjotarora, line 44: It could be useful to have an additional `expirable` field of type [`driver.Expirable](https://github.com/mongodb/mongo-go-driver/blob/90a0db4834452f0d85433bfef1a1799318384cec/x/mongo/driver/driver.go#L64). You should be able to directly cast `conn` to a `driver.Expirable` without any issues.

proxy.go
```diff
@@ -1,62 +1,87 @@
...
-	if me.err != nil {
-		doc = append(doc, bson.E{"errmsg", me.err.Error()})
+func (m *MongoConnectionWrapper) Close() {
+	if m.conn == nil {
+		logTrace(m.logger, m.trace, "mongo connection is nil!")
+		return
+	}
+	id := m.conn.ID()
+	if m.bad {
+		if m.conn != nil && m.conn.ID() != "<closed>" {
```
divjotarora, line 58: Instead of checking for the special `<closed>` ID, you can cast to `driver.Expirable` and use the `Alive` function.

proxy.go
```diff
@@ -1,62 +1,87 @@
...
+			logTrace(m.logger, m.trace, "closing underlying bad mongo connection %v", id)
+			ec, ok := m.conn.(*topology.Connection)
```
divjotarora, line 60: Use `driver.Expirable` instead of `*topology.Connection`.

proxy.go
```diff
@@ -1,62 +1,87 @@
...
+			}
+			if err := ec.Expire(); err != nil {
+				logTrace(m.logger, m.trace, "failed to expire connection. %v", err)
+			}
+		} else {
+			logTrace(m.logger, m.trace, "bad mongo connection is nil!")
+		}
+	}
+	logTrace(m.logger, m.trace, "closing mongo connection %v", id)
+	if m.conn.ID() != "<closed>" {
```
divjotarora, line 72: Same comment here about casting to `driver.Expirable` and using the `Alive` method.

proxy.go
```diff
@@ -187,127 +209,195 @@ func (ps *ProxySession) respondWithError(clientMessage Message, err error) error
...
-	}
-	mongoConn := pooledConn.conn
-
-	err = SendMessage(m, mongoConn)
+	// Send message to mongo
+	err = mongoConn.conn.WriteWireMessage(ps.proxy.Context, m.Serialize())
 	if err != nil {
-		pooledConn.bad = true
-		return pooledConn, NewStackErrorf("error writing to mongo: %v", err)
+		mongoConn.bad = true
```
divjotarora, line 294: You might not need to mark the connection as "bad" here. The driver will close the socket when it gets a socket error during `WriteWireMessage`, so there's no need to do an additional explicit `Expire` call in `MongoConnectionWrapper.Close()`. Connections can be closed multiple times, though, so it's not incorrect to have this. The `Expire` call will acquire a mutex, but it should be uncontended and therefore shouldn't block, so this might be a micro-optimization to think about later on.

proxy.go
TomerYakir: yeah - handling this now

mongo_error.go
```diff
@@ -0,0 +1,40 @@
...
+	code     int
+	codeName string
+}
+
+func NewMongoError(err error, code int, codeName string) MongoError {
+	return MongoError{err, code, codeName}
+}
+
+func (me MongoError) ToBSON() bson.D {
+	doc := bson.D{{"ok", 0}}
```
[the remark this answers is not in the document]
TomerYakir: I'm not 100% sure if the order matters, but as far as I've seen, mongo always returned the fields in that order so keeping this as is.
btw, this method wasn't actually changed as part of this PR, just moved around.

proxy.go
TomerYakir: yeah, I'm not sure why were doing that either. I'll see if changing this doesn't break anything ¯\_(ツ)_/¯

proxy.go
TomerYakir: good catch! changed.

proxy.go
TomerYakir: that was a mistake - changed!

proxy.go
TomerYakir: I was following Louisa's comment here - https://github.com/mongodb/mongonet/compare/drivertest#diff-ddf6a9e7c9b2d4f21e3a3eabfe7c7d77R289 - but might have misinterpreted her intention?

divjotarora: In that case, I'm fine with having them be separate. It might be possible to combine some the code duplicated between `ReadMessage` and `ReadMessageFromBytes` to clean things up. For example, the size validation can probably be delegated to a helper function and all of the code that happens after `ReadMessage` reads the remaining bytes into `restBuf` can be shared between the two.

proxy.go
TomerYakir: cool - btw - is the same true for `ReadWireMessage`?

divjotarora: Yeah, it should be the same. Also, forgot to ask in my original comment, but do we have tests for these cases? If not, it might be worth trying to add those if we're considering this optimization.

proxy.go
TomerYakir: `conn` is of type `driver.Connection`. Did you mean I should be able to do `driver.Expirable(conn)`?
I'm getting an error:
```
cannot convert conn (type driver.Connection) to type driver.Expirable:
	driver.Connection does not implement driver.Expirable (missing Alive method)
```

proxy.go
TomerYakir: We don't have explicit tests for that, but our application int tests have a "hostile" conditions section that's bound to hit those, but I'll look into adding some tests to mongonet to simulate errors on ReadWireMessage and WriteWireMessage.

proxy.go
divjotarora: Can you do `conn.(driver.Expirable)`?

TomerYakir: derp. yes

proxy.go
```diff
@@ -187,97 +208,129 @@ func (ps *ProxySession) respondWithError(clientMessage Message, err error) error
...
 func (ps *ProxySession) Close() {
 	ps.interceptor.Close()
 }
 
-func (ps *ProxySession) doLoop(pooledConn *PooledConnection) (*PooledConnection, error) {
+func logPanic(logger *slogger.Logger) {
+	if r := recover(); r != nil {
+		var stacktraces bytes.Buffer
+		pprof.Lookup("goroutine").WriteTo(&stacktraces, 2)
+		logger.Logf(slogger.ERROR, "Recovering from mongonet panic. error is: %v \n stack traces: %v", r, stacktraces.String())
```
tolsen, line 223: Please logger.Flush() before panicking

server_test.go
tolsen: TODO is still here

tolsen (changes requested):

proxy.go
TomerYakir: fixed

tolsen (approved):

iwysiu (approved):

proxy.go
```diff
@@ -187,97 +208,130 @@ func (ps *ProxySession) respondWithError(clientMessage Message, err error) error
...
 		if m == nil {
 			// already responded
-			return pooledConn, nil
+			return mongoConn, nil
 		}
 	}
-
-	if pooledConn == nil {
-		pooledConn, err = ps.proxy.connPool.Get()
+	if mongoConn == nil || !mongoConn.expirableConn.Alive() {
```
divjotarora, line 284: Do we expect there to be cases where `DoLoopTemp` passed in a `mongoConn` value that isn't alive or is this here as a defensive check?

divjotarora (commented):
Looks good! Just some remaining questions about edge cases and when we expect them to happen.

divjotarora:
@TomerYakir Forgot to mention this in my last review, but if we're changing the behavior to not mark the proxy to MongoDB connection as "bad" if the `ReadWireMessage` or `WriteWireMessage` calls fail, I think we should figure out ways to unit test that inside mongonet. If you agree, feel free to reach out in the Slack channel and we can brainstorm ways to make those functions fail.

proxy.go
```diff
@@ -187,97 +208,130 @@ func (ps *ProxySession) respondWithError(clientMessage Message, err error) error
...
+		pprof.Lookup("goroutine").WriteTo(&stacktraces, 2)
+		logger.Logf(slogger.ERROR, "Recovering from mongonet panic. error is: %v \n stack traces: %v", r, stacktraces.String())
+		logger.Flush()
+		panic(r)
+	}
+}
+
+func extractTopology(mc *mongo.Client) *topology.Topology {
+	e := reflect.ValueOf(mc).Elem()
+	d := e.FieldByName("deployment")
```
kevinAlbs, line 231: IIUC this is reaching inside `mongo.Client` to grab the private `deployment` field. I assume this is all ok for now, until GODRIVER-1760 is implemented. Should this be checking for the zero value? If `deployment` were renamed, I'm not sure what type of error would occur.

sock.go
```diff
@@ -20,6 +20,65 @@ func sendBytes(writer io.Writer, buf []byte) error {
 
 }
 
+func getMessage(header MessageHeader, body []byte) (Message, error) {
+	switch header.OpCode {
+	case OP_REPLY:
```
kevinAlbs, line 25: I assume this is for completeness in mongonet, but the only op codes Atlas Proxy needs to support are `OP_QUERY`, `OP_REPLY`, and `OP_MSG`, right?

sock.go
```diff
@@ -20,6 +20,65 @@ func sendBytes(writer io.Writer, buf []byte) error {
...
+		return parseCommandReplyMessage(header, body)
+	case OP_MSG:
+		return parseMessageMessage(header, body)
+	default:
+		return nil, NewStackErrorf("unknown op code: %v", header.OpCode)
+	}
+}
+
+func validateHeaderSize(headerSize int32) error {
+	if headerSize > int32(200*1024*1024) {
```
kevinAlbs, line 51: It looks like this was moved, but I'm curious if there is a reason behind 200MB. I'd think `maxMessageSizeBytes` would be the number to check.

sock.go
divjotarora: I don't think that's true. IIUC, there will essentially be two modes for the proxy: one for shared and one for serverless. Seems like we're making the assumption that the serverless proxy will only have to support 5.0+, so it only needs to support those three opcodes (though the exact decision about that is pending Louisa's opinion on the CLOUDP ticket about bumping minWireVersion). I'm guessing the shared tier proxy still needs to continue supporting all of these opcodes for backwards compatibility, though none of our drivers would send them.

proxy.go
divjotarora: If the `deployment` field doesn't exist, `FieldByName` will return a zero-valued `reflect.Value`  and the `d.Type()` call on the next line will panic. I'm fine with this behavior, but I agree that adding an explicit `d.IsZero` check would make things more explicit.

proxy.go
```diff
@@ -1,62 +1,86 @@
...
-	if me.err != nil {
-		doc = append(doc, bson.E{"errmsg", me.err.Error()})
+func (m *MongoConnectionWrapper) Close() {
+	if m.conn == nil {
+		logTrace(m.logger, m.trace, "mongo connection is nil!")
+		return
+	}
+	id := m.conn.ID()
+	if m.bad {
+		if m.conn != nil && m.expirableConn.Alive() {
```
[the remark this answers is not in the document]
TomerYakir: This is definitely redundant at this point (see the condition on line 55) so I removed it.
As for the condition on line 55 - yes, it's just defensive programming and I bumped the log level so we can further look into if that ever happens.

proxy.go
TomerYakir: this is just a defensive check.

cmd/straight_proxy/straight_proxy.go
```diff
@@ -15,15 +15,18 @@ func main() {
...
 	pc.MongoSSLSkipVerify = true
 
-	proxy := mongonet.NewProxy(pc)
+	proxy, err := mongonet.NewProxy(pc)
+	if err != nil {
+		panic(err)
+	}
 
 	proxy.InitializeServer()
 	proxy.OnSSLConfig(nil)
```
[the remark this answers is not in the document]
TomerYakir: added a check

sock.go
TomerYakir: yeah, that was moved and I'm unsure about that figure. I'll file a separate ticket to look into that value.

TomerYakir:
@divjotarora tests added - RFAL

proxy_test.go
```diff
@@ -0,0 +1,270 @@
...
+		}
+		mm.Query = qb
+
+		return mm, nil, nil
+	}
+
+	return m, nil, nil
+}
+
+func GetConnCreated(p *Proxy) int64 {
```
divjotarora, line 79: nit: If the caller has access to a `Proxy` instance, can they directly call `p.GetConnectionsCreated()` rather than going through this function?

proxy_test.go
```diff
@@ -0,0 +1,270 @@
...
+	return p.GetConnectionsCreated()
+}
+
+func doFind(proxyPort, iteration int, shouldFail bool) error {
+	opts := options.Client().ApplyURI(fmt.Sprintf("mongodb://localhost:%d", proxyPort))
```
divjotarora, line 84: Seems like the code to create a `Client` could be shared by this function and `enableFailPoint`/`disableFailPoint`.

proxy_test.go
```diff
@@ -0,0 +1,270 @@
...
+	ctx, cancelFunc := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancelFunc()
+	if err := client.Connect(ctx); err != nil {
+		return fmt.Errorf("cannot connect to server. err: %v", err)
+	}
+	defer client.Disconnect(ctx)
+	cmd := bson.D{
+		{"configureFailPoint", "failCommand"},
+		{"mode", "alwaysOn"},
+		// {"mode", bson.D{{"times", 5}}},
```
divjotarora, line 143: nit: remove this commented-out line

proxy_test.go
```diff
@@ -0,0 +1,270 @@
...
+		{"data", bson.D{
+			{"failCommands", []string{"find"}},
+			{"closeConnection", true},
+		}},
+	}
+	res := client.Database("admin").RunCommand(ctx, cmd)
```
divjotarora, line 149: This should probably be
```
return client.Database("admin").RunCommand(ctx, cmd).Err()
```

I'm assuming the `Println` below is leftover debugging. If it's there intentionally, you can instead do
```
res, err := client.Database("admin").RunCommand(ctx, cmd).DecodeBytes()
fmt.Println(res)
return err
```

proxy_test.go
```diff
@@ -0,0 +1,270 @@
...
+	defer cancelFunc()
+	if err := client.Connect(ctx); err != nil {
+		return fmt.Errorf("cannot connect to server. err: %v", err)
+	}
+	defer client.Disconnect(ctx)
+	cmd := bson.D{
+		{"configureFailPoint", "failCommand"},
+		{"mode", "off"},
+	}
+	res := client.Database("admin").RunCommand(ctx, cmd)
```
divjotarora, line 171: Same comment here about using either `Err` or `DecodeBytes`.

proxy_test.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+func runFinds(proxyPort int, shouldFail bool) int32 {
+	var wg sync.WaitGroup
+	var failing int32
+	for i := 0; i < 5; i++ {
+		wg.Add(1)
+		go func() {
+			err := doFind(proxyPort, i, shouldFail)
+			wg.Done()
```
divjotarora, line 183: nit: The wait group `Done` call is generally done using `defer wg.Done()` as the first call of the goroutine. This ensures that the wait group is decremented even if another call in the goroutine panics.

divjotarora (commented):
All nits/minor comments in the new tests. Correctness-wise everything LGTM.

divjotarora (approved):
LGTM!

PR #56, opened 2020-10-14
Go driver connection pool server selection

TomerYakir:
This adds server selection and getting read preference from the client message.
Note I had to revise the int tests a bit as enabling failPoint on "find" when replica sets are involved has broken some driver machinery (transactions).
Also I tweaked the connection count checking on the tests so it'll be less sensitive.


proxy.go
```diff
@@ -59,10 +65,10 @@ func logMessageTrace(logger *slogger.Logger, trace bool, m Message) {
 			doc, _ = mm.Docs[0].ToBSOND()
 		default:
 			// not bothering about printing other message types
-			return
+			msg = fmt.Sprintf("got another type %T", mm)
```
louisaberger, line 68: It looks like this gets immediately overwritten on line 71?

proxy.go
```diff
@@ -265,10 +271,66 @@ func extractTopology(mc *mongo.Client) *topology.Topology {
...
 
-func (ps *ProxySession) getMongoConnection() (*MongoConnectionWrapper, error) {
-	srv, err := ps.proxy.topology.FindServer(ps.proxy.descriptionServer)
-	if err != nil {
-		return nil, err
+func getReadPrefFromOpMsg(mm *MessageMessage) (rp *readpref.ReadPref) {
+	for _, section := range mm.Sections {
+		bs, ok := section.(*BodySection)
+		if !ok {
+			continue
```
louisaberger, line 278: Is there any chance that we never reach a `BodySection` in the `Sections`?

proxy.go
```diff
@@ -265,10 +271,66 @@ func extractTopology(mc *mongo.Client) *topology.Topology {
...
+		}
+		bsd := bsoncore.Document(bs.Body.BSON)
```
louisaberger, line 280: Can `bs.Body` ever be nil? (would cause the .BSON lookup to panic)

proxy.go
```diff
@@ -265,10 +271,66 @@ func extractTopology(mc *mongo.Client) *topology.Topology {
...
+			// not concerned about the error - we'll just fallback to readpref=primary
+			return
+		}
+		rpDoc, ok := rpVal.DocumentOK()
+		if !ok {
+			return
+		}
+		opts := make([]readpref.Option, 0, 1)
+		if maxStalenessVal, err := rpDoc.LookupErr("maxStalenessSeconds"); err == nil {
+			if maxStalenessVal.IsNumber() {
```
louisaberger, line 292: Do we want to log something if maxStalenessVal.IsNumber() returns false? When could that happen?

proxy.go
```diff
@@ -315,11 +389,11 @@ func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper) (*MongoConnect
 		}
 	}
 	if mongoConn == nil || !mongoConn.expirableConn.Alive() {
-		mongoConn, err = ps.getMongoConnection()
+		mongoConn, err = ps.getMongoConnection(rp)
 		if err != nil {
-			return nil, NewStackErrorf("cannot get connection to mongo %v", err)
+			return nil, NewStackErrorf("cannot get connection to mongo %v using connection mode %v", err, ps.proxy.config.ConnectionMode)
```
louisaberger, line 394: Perhaps we also want to log the rp here?

proxy.go
```diff
@@ -371,29 +446,29 @@ func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper) (*MongoConnect
...
+			return mongoConn, NewStackErrorf("bad response type from server %T", r)
 		}
 	}
 }
 
 func NewProxy(pc ProxyConfig) (Proxy, error) {
 	ctx := context.Background()
 	var initCount int64 = 0
-	p := Proxy{pc, nil, nil, nil, nil, description.Server{Addr: address.Address(pc.MongoAddress())}, ctx, &initCount}
+	p := Proxy{pc, nil, nil, nil, nil, description.Server{Addr: address.Address(pc.MongoAddress())}, readpref.Primary(), ctx, &initCount}
```
louisaberger, line 471: Why is this always set to Primary?

proxy_test.go
```diff
@@ -185,21 +332,88 @@ func runFinds(proxyPort int, shouldFail bool, t *testing.T) int32 {
 	return failing
 }
 
-// backing mongo must be started with --setParameter enableTestCommands=1
-func TestProxySanity(t *testing.T) {
+func getProxyConfigForMongod(hostname string, mongoPort, proxyPort int) ProxyConfig {
+	pc := NewProxyConfig("localhost", proxyPort, "", hostname, mongoPort, "", "", "test proxy", true, Direct, 5)
```
louisaberger, line 336: You may want to consider defining a constant for this testing value of "5" that's used in a few different files

proxy_test.go
```diff
@@ -185,21 +332,88 @@ func runFinds(proxyPort int, shouldFail bool, t *testing.T) int32 {
...
+	if err := disableFailPoint("localhost", mongoPort, Direct); err != nil {
+		t.Fatalf("failed to disable failpoint. err=%v", err)
+		return
+	}
+	pc := getProxyConfigForMongod("localhost", mongoPort, proxyPort)
+	privateTester(t, pc, "localhost", proxyPort, mongoPort, 5, Direct, false)
+}
+
+// this is expected to go through the same code as primary mode since Mongod mode essentially ignores read preference from client
+func TestProxySanityMongodModeSecondary(t *testing.T) {
```
louisaberger, line 357: Code looks the same here except for the false/true in privateTester -- could combine code?

proxy_test.go
```diff
@@ -185,21 +332,88 @@ func runFinds(proxyPort int, shouldFail bool, t *testing.T) int32 {
...
+	if err := disableFailPoint("localhost", mongoPort, Direct); err != nil {
 		t.Fatalf("failed to disable failpoint. err=%v", err)
 		return
 	}
-	pc := NewProxyConfig("localhost", proxyPort, "localhost", mongoPort, "", "", "test proxy", true)
+	pc := getProxyConfigForMongod("localhost", mongoPort, proxyPort)
+	privateTester(t, pc, "localhost", proxyPort, mongoPort, 5, Direct, true)
+}
+
+func getProxyConfigForMongos(hostname string, mongoPort, proxyPort int) ProxyConfig {
```
louisaberger, line 371: Same thought with code repeats here

proxy_test.go
```diff
@@ -185,21 +332,88 @@ func runFinds(proxyPort int, shouldFail bool, t *testing.T) int32 {
...
+	uri := fmt.Sprintf("mongodb://%s:%v,%s:%v,%s:%v/?replSet=proxytest", hostname, mongoPort, hostname, mongoPort+1, hostname, mongoPort+2)
+	pc := NewProxyConfig("localhost", proxyPort, uri, hostname, mongoPort, "", "", "test proxy", true, Cluster, 5)
 	pc.MongoSSLSkipVerify = true
-	pc.InterceptorFactory = &MyFactory{}
+	pc.InterceptorFactory = &MyFactory{Cluster, mongoPort, proxyPort}
+	return pc
+}
+
+func TestProxySanityMongosModePrimary(t *testing.T) {
```
louisaberger, line 379: These also look the same except for the bool

proxy_test.go
```diff
@@ -211,55 +425,65 @@ func TestProxySanity(t *testing.T) {
...
 	if atomic.LoadInt32(&failing) > 0 {
-		t.Fatalf("finds failures")
+		t.Fatalf("ops failures")
 		return
 	}
 
-	if conns := proxy.GetConnectionsCreated(); conns != 10 {
-		t.Fatalf("expected connections created to equal 10 but was %v", conns)
+	conns = proxy.GetConnectionsCreated()
+	if conns == currConns {
```
louisaberger, line 468: Might want to check for conns >= currConns for completeness (and below on 486)

louisaberger (commented):
Nice! Mostly comments around adding logging safeguards, and some testing code abstraction

proxy.go
TomerYakir: not in valid OP_MSG, but in theory it's possible. In this case, we'll simply fallback to the default read preference (primary).

proxy.go
TomerYakir: I don't think so. `Body` is a field of type `SimpleBSON` on the `BodySection` struct and `BSON` is a field on the `SimpleBSON` struct.
Having said that, I added a safeguard just in case `bs` is somehow nil.

proxy.go
TomerYakir: Added logging.
I guess a low-level client might send malformed values for maxStalenessSeconds so this was added as a safeguard.

proxy.go
TomerYakir: that's just the default - refactored a bit to make that clearer

proxy_test.go
TomerYakir: The full snippet is this:
```
if conns == currConns {
	t.Fatalf("expected connections created to increase from (%v), but got %v", currConns, conns)
}
```
connectionsCreated cannot decrease. if they stay the same the test should fail. if they increase it's fine. should I revise the error message?

proxy.go
```diff
@@ -46,23 +51,33 @@ func logTrace(logger *slogger.Logger, trace bool, format string, args ...interfa
...
 		var doc bson.D
+		var msg string
+		var err error
 		switch mm := m.(type) {
 		case *MessageMessage:
+		SECTIONS:
 			for _, section := range mm.Sections {
 				if bs, ok := section.(*BodySection); ok {
-					doc, _ = bs.Body.ToBSOND()
+					doc, err = bs.Body.ToBSOND()
```
divjotarora, line 61: If we're changing this to handle an error, should we also change the other cases to match?

proxy.go
divjotarora: Agreed that this isn't possible in a valid OP_MSG, but I think we should be stricter about it when it happens rather than falling back to the default. Maybe that needs to happen higher up the stack when we actually parse the incoming message, though, so we can make assumptions about message structure without worrying about every edge case when we get here. This comment also applies to the `if bs == nil` check below.

proxy.go
divjotarora: An actual mongos will error if the `$readPreference` field is not a document or if any of the fields in the document are not of the expected type. I think it would make sense to mimic that behavior rather than just logging and silently overwriting to the default.

proxy.go
```diff
@@ -265,10 +280,85 @@ func extractTopology(mc *mongo.Client) *topology.Topology {
...
+			return
+		}
+		rpDoc, ok := rpVal.DocumentOK()
+		if !ok {
+			logger.Logf(slogger.WARN, "$readPreference isn't a document. falling back to default read preference")
+			return
+		}
+		opts := make([]readpref.Option, 0, 1)
+		if maxStalenessVal, err := rpDoc.LookupErr("maxStalenessSeconds"); err == nil {
+			if maxStalenessVal.IsNumber() {
```
divjotarora, line 308: style nit: You can skip the `IsNumber` check and just do `AsInt32OK` instead to do the type checking and int32 conversion in one function call.

proxy.go
```diff
@@ -265,10 +280,85 @@ func extractTopology(mc *mongo.Client) *topology.Topology {
...
+			case "primarypreferred":
+				return readpref.PrimaryPreferred(opts...)
+			case "secondary":
+				return readpref.Secondary(opts...)
+			case "secondarypreferred":
+				return readpref.SecondaryPreferred(opts...)
+			case "nearest":
+				return readpref.Nearest(opts...)
+			default:
+				return
```
divjotarora, line 333: We should probably error if we get a mode we don't expect.

proxy.go
```diff
@@ -340,24 +442,25 @@ func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper) (*MongoConnect
...
 			if err == io.EOF {
 				return nil, err
 			}
 			return nil, NewStackErrorf("got error reading response from mongo %v", err)
 		}
-		logMessageTrace(ps.proxy.logger, ps.proxy.config.TraceConnPool, resp)
 		if respInter != nil {
 			resp, err = respInter.InterceptMongoToClient(resp)
 			if err != nil {
-				return nil, NewStackErrorf("error intercepting message %v", err)
```
divjotarora, line 357: Why the change from `return nil` to `return mongoConn` both here and in some other statements below?

proxy.go
```diff
@@ -416,7 +520,15 @@ func getMongoClient(p *Proxy, pc ProxyConfig, ctx context.Context) (*mongo.Clien
 					p.AddConnection()
 				}
 			},
-		})
+		}).
+		SetServerSelectionTimeout(time.Duration(pc.ServerSelectionTimeoutSec) * time.Second)
+
+	if pc.ConnectionMode == Direct {
```
divjotarora, line 526: Should we consider calling `ApplyURI` before setting additional options? I don't think it's a problem right now, but if the URI had a query parameter like `directConnection=false`, that would override the `SetDirect()` call above. I'm assuming we want the `SetDirect` call to override, though.

proxy.go
TomerYakir: ugh.. merge issue - will fix!

proxy.go
TomerYakir: moved the check to a higher level on the stack

proxy.go
```diff
@@ -265,10 +280,85 @@ func extractTopology(mc *mongo.Client) *topology.Topology {
...
+	}
+	return
+}
+
+func (ps *ProxySession) getMongoConnection(rp *readpref.ReadPref) (*MongoConnectionWrapper, error) {
+	var err error
+	var srv driver.Server
+	var found bool
+	if ps.proxy.config.ConnectionMode == Direct {
+		srv2, err := ps.proxy.topology.FindServer(ps.proxy.descriptionServer)
```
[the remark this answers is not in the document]
TomerYakir: I might be misunderstanding your meaning, but are you suggesting to do something like this?
```
var srvSelector description.ServerSelector
if ps.proxy.config.ConnectionMode == Direct {
    srvSelector = ... // something that returns a single candidate
} else {
    srvSelector = description.ReadPrefSelector(rp)
}
srv, err = ps.proxy.topology.SelectServer(ps.proxy.Context, srvSelector)
		if err != nil {
			return nil, err
		}
```

If so, isn't `topology.SelectServer`call `FindServer` under the hood?
```
selected := suitable[rand.Intn(len(suitable))]
selectedS, err := t.FindServer(selected)
```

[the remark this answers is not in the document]
TomerYakir: oh I think you meant to stop using `SelectServer` - working on that now.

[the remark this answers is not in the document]
TomerYakir: I think I'll need some further guidance on this as my current approaches either do not avoid the locks or appear to be bypassing driver mechanisms.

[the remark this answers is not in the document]
divjotarora: Sorry, I forgot that `SelectServer` internally does call `FindServer`, so it is not a lock-free process. And yes, I am suggesting that you do something like

```
var srvSelector description.ServerSelector
if ps.proxy.config.ConnectionMode == Direct {
    srvSelector = ... // something that returns a single candidate
} else {
    srvSelector = description.ReadPrefSelector(rp)
}
srv, err = ps.proxy.topology.SelectServer(ps.proxy.Context, srvSelector)
```

Because the ability to select a server given its address will also be useful for other features like cursors, you could consider having a helper function like `func createPinnedServerSelector(addr address.Address) description.ServerSelector` (see [description.ReadPrefSelector](https://github.com/mongodb/mongo-go-driver/blob/935fe3f869a9d46940f402873c97c19d19a8d41d/mongo/description/server_selector.go#L128)) for an idea of the implementation.

proxy.go
louisaberger: Cool! I'm not seeing that here.. is that pushed up?

proxy_test.go
louisaberger: Could this code and the `getProxyConfigForMongod` still be abstracted with a common function?

TomerYakir:
@louisaberger @divjotarora all comments addressed - RFAL

proxy.go
```diff
@@ -265,10 +280,85 @@ func extractTopology(mc *mongo.Client) *topology.Topology {
...
+			if maxStalenessVal.IsNumber() {
+				maxStalenessSec := maxStalenessVal.AsInt32()
+				if maxStalenessSec > 0 {
+					opts = append(opts, readpref.WithMaxStaleness(time.Duration(maxStalenessSec)*time.Second))
+				}
+			} else {
+				logger.Logf(slogger.WARN, "maxStalenessSeconds %v isn't a number. ignoring it", maxStalenessVal)
+			}
+		}
+		if modeVal, err := rpDoc.LookupErr("mode"); err == nil {
```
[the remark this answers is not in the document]
divjotarora: Seems like we're still allowing `$readPreference` documents without a `mode` field?

proxy.go
```diff
@@ -265,8 +287,76 @@ func extractTopology(mc *mongo.Client) *topology.Topology {
...
+			}
+			return nil, fmt.Errorf("got an error looking up $readPreference: %v", err)
+		}
+		rpDoc, ok := rpVal.DocumentOK()
+		if !ok {
+			return nil, fmt.Errorf("$readPreference isn't a document")
+		}
+		opts := make([]readpref.Option, 0, 1)
+		if maxStalenessVal, err := rpDoc.LookupErr("maxStalenessSeconds"); err == nil {
+			if maxStalenessSec, ok := maxStalenessVal.AsInt32OK(); ok && maxStalenessSec > 0 {
```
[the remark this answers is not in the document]
TomerYakir: changed to error.
I thought I saw somewhere that the value has to be greater than 0 (or even 90).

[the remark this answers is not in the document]
divjotarora: It's probably worth figuring out what an actual mongos does here and mimicking that. Feel free to resolve this conversation if you think the existing checks are enough, though.

proxy.go
```diff
@@ -332,6 +332,8 @@ func getReadPrefFromOpMsg(mm *MessageMessage, logger *slogger.Logger, defaultRp
 			default:
 				return nil, fmt.Errorf("got unsupported read preference %v", modeStr)
 			}
+		} else {
+			return nil, fmt.Errorf("error looking up the 'mode' field in read preference: %v", err2)
```
divjotarora, line 336: We probably don't want to leak errors from the Go Driver out of mongonet. The error will just be that the field doesn't exist, so you can probably do `return nil, errors.New("read preference is missing the required \"mode\" field")`.

iwysiu (approved):

proxy.go
```diff
@@ -46,23 +51,41 @@ func logTrace(logger *slogger.Logger, trace bool, format string, args ...interfa
 func logMessageTrace(logger *slogger.Logger, trace bool, m Message) {
 	if trace {
 		var doc bson.D
+		var msg string
+		var err error
 		switch mm := m.(type) {
 		case *MessageMessage:
+		SECTIONS:
```
divjotarora, line 58: nit: Why do we need a new label? These are generally used for nested loops when you want to break out of the outer loop, but there's only one loop here. I think a regular `break` statement would be sufficient.

divjotarora (approved):
Small nit, but otherwise LGTM.

proxy.go
TomerYakir: not sure why I added that. removed!

TomerYakir:
Waiting for @louisaberger to approve before I merge this in.

louisaberger (approved):

PR #57, opened 2020-10-15
BSON lib optimizations

TomerYakir:


divjotarora (approved):
LGTM as long as there aren't any direct invocations of `bson.Marshal` and `bson.Unmarshal` in mongonet.

TomerYakir:
these are the only references to bson.Marshal and bson.Unmarshal in mongonet

PR #58, opened 2020-11-05
Handle SDAM errors

TomerYakir:
When drivers are connected to a cluster, they need to make sure they don’t route operations to nodes that are down. They handle this by:

1. Periodic background monitoring checks - This will be handled by our mongo.Client
2. Application error handling:
  - Drivers have some error handling code that marks a node as “down” if needed.
  - We will need to handle this part because we will be bypassing the driver’s operations layer

### DIFF
--- a/proxy.go
+++ b/proxy.go
@@ -93,6 +93,7 @@ func logMessageTrace(logger *slogger.Logger, trace bool, m Message) {
 // MongoConnectionWrapper is used to wrap the driver connection so we can explicitly expire (close out) connections on certain occasions that aren't being picked up by the driver
 type MongoConnectionWrapper struct {
 	conn          driver.Connection
+	srv           driver.Server
 	expirableConn driver.Expirable
 	bad           bool
 	logger        *slogger.Logger
@@ -378,7 +379,12 @@ func (ps *ProxySession) getMongoConnection(rp *readpref.ReadPref) (*MongoConnect
 	if !ok {
 		return nil, fmt.Errorf("bad connection type %T", conn)
 	}
-	return &MongoConnectionWrapper{conn, ec, false, ps.proxy.logger, ps.proxy.config.TraceConnPool}, nil
+	return &MongoConnectionWrapper{conn, srv, ec, false, ps.proxy.logger, ps.proxy.config.TraceConnPool}, nil
+}
+
+func wrapNetworkError(err error) error {
+	labels := []string{driver.NetworkError}
+	return driver.Error{Message: err.Error(), Labels: labels, Wrapped: err}
 }
 
 func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper) (*MongoConnectionWrapper, error) {
@@ -438,9 +444,15 @@ func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper) (*MongoConnect
 		logTrace(ps.proxy.logger, ps.proxy.config.TraceConnPool, "got new connection %v using connection mode=%v readpref=%v", mongoConn.conn.ID(), ps.proxy.config.ConnectionMode, rp)
 	}
 
+	ep, ok := mongoConn.srv.(driver.ErrorProcessor)
+	if !ok {
+		return nil, NewStackErrorf("server ErrorProcessor type assertion failed")
+	}
+
 	// Send message to mongo
 	err = mongoConn.conn.WriteWireMessage(ps.proxy.Context, m.Serialize())
 	if err != nil {
+		ep.ProcessError(wrapNetworkError(err), mongoConn.conn)
 		return mongoConn, NewStackErrorf("error writing to mongo: %v", err)
 	}
 
@@ -457,6 +469,7 @@ func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper) (*MongoConnect
 		ret, err := mongoConn.conn.ReadWireMessage(ps.proxy.Context, nil)
 		if err != nil {
 			logTrace(ps.proxy.logger, ps.proxy.config.TraceConnPool, "error reading wire message mongo conn %v %v", mongoConn.conn.ID(), err)
+			ep.ProcessError(wrapNetworkError(err), mongoConn.conn)
 			return nil, NewStackErrorf("error reading wire message from mongo conn %v: %v", mongoConn.conn.ID(), err)
 		}
 		logTrace(ps.proxy.logger, ps.proxy.config.TraceConnPool, "read data from mongo conn %v", mongoConn.conn.ID())
--- a/proxy.go
+++ b/proxy.go
@@ -500,6 +500,7 @@ func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper) (*MongoConnect
 		err = SendMessage(resp, ps.conn)
 		if err != nil {
 			mongoConn.bad = true
+			logTrace(ps.proxy.logger, ps.proxy.config.TraceConnPool, "got error sending response to client from conn %v: %v", mongoConn.conn.ID(), err)
 			return nil, NewStackErrorf("got error sending response to client from conn %v: %v", mongoConn.conn.ID(), err)
 		}
 		logTrace(ps.proxy.logger, ps.proxy.config.TraceConnPool, "sent back data to user from mongo conn %v", mongoConn.conn.ID())
--- a/proxy.go
+++ b/proxy.go
@@ -39,6 +39,7 @@ type Proxy struct {
 
 	Context            context.Context
 	connectionsCreated *int64
+	poolCleared        *int64
 }
 
 func logTrace(logger *slogger.Logger, trace bool, format string, args ...interface{}) {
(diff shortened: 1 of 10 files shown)


TomerYakir:
Hi @divjotarora - I want to make sure I fully understand the goal here.
On the current state, if the proxy goes down, the driver will correctly update the topology marking the server as unknown.
If the underlying mongo goes down however, the driver won't update the topology error unless we explicitly call `processError` so we need to intercept errors when reading and writing messages to the underlying mongo so that the proxy->mongo connection view gets updated. Is that correct?

divjotarora:
@TomerYakir Yeah, that's correct. `ProcessError` is required to make sure we're updating the driver's internal state of the cluster synchronously. If we don't call it, we could continue to target stale/down servers until the background monitoring routine picks up the state change. This was a much bigger risk in 4.2 and below because the background monitoring routine only picked up changes once every 10 seconds, so we could potentially use a stale view of the cluster for server selection for up to 10s. In 4.4+, the background routine uses a pub/sub model so the server publishes state changes as they happen and we pick them up much more quickly in drivers. However, `ProcessError` is still necessary for correctness, especially if we implement retryability in the proxy.

TomerYakir:
@divjotarora this is now ready for review. Added tests based on PoolCleared as discussed on slack.

louisaberger (approved):
LGTM from our perspective, but mainly deferring to @divjotarora for the expected behavior from a drivers perspective

proxy.go
```diff
@@ -93,6 +94,7 @@ func logMessageTrace(logger *slogger.Logger, trace bool, m Message) {
 // MongoConnectionWrapper is used to wrap the driver connection so we can explicitly expire (close out) connections on certain occasions that aren't being picked up by the driver
 type MongoConnectionWrapper struct {
 	conn          driver.Connection
+	srv           driver.Server
```
divjotarora, line 97: Can we instead store a `driver.ErrorProcessor` and change `getMongoConnection` to do the cast when the connection is created? That would avoid re-casting each time we call `doLoop` with the same connection.

divjotarora (commented):
Couple of larger comments:

1. This is handling network errors on reads/writes, but is not handling state change errors that are returned via well-formed response documents (e.g. an `InterruptedAtShutdown` error returned as `{ok: 0, code: 11660}`). You'll likely have to call the driver's [`extractError` function](https://github.com/mongodb/mongo-go-driver/blob/21d9c44f1bb8bca1fe8c0b98adda469f388ed6e7/x/mongo/driver/errors.go#L330) on the response document and use use that to call `ProcessError` again. I realize that this function is currently unexported, so you can't use it directly. To unblock testing, you can copy over this function for now. I've filed GODRIVER-1784 to export it so it can be used directly.

2. The tests all assert that `poolCleared` is 0. If there are state change errors, though, the pool should be cleared. Are the tests not correctly reproducing network errors?

TomerYakir:
> Couple of larger comments:
> 
> 1. This is handling network errors on reads/writes, but is not handling state change errors that are returned via well-formed response documents (e.g. an `InterruptedAtShutdown` error returned as `{ok: 0, code: 11660}`). You'll likely have to call the driver's [`extractError` function](https://github.com/mongodb/mongo-go-driver/blob/21d9c44f1bb8bca1fe8c0b98adda469f388ed6e7/x/mongo/driver/errors.go#L330) on the response document and use use that to call `ProcessError` again. I realize that this function is currently unexported, so you can't use it directly. To unblock testing, you can copy over this function for now. I've filed GODRIVER-1784 to export it so it can be used directly.
> 2. The tests all assert that `poolCleared` is 0. If there are state change errors, though, the pool should be cleared. Are the tests not correctly reproducing network errors?

1. To confirm my understanding - in those cases `ReadWireMessage` wouldn't return an error and I'll have to cast the response to `bsoncore.Document`, then pass it through `extractError` and if that returns an error I'll process that error using `ProcessError`. Correct?

2. Actually we're checking for non-0 values for poolCleared - See proxy_test:489 on this feature branch.

divjotarora:
> To confirm my understanding - in those cases ReadWireMessage wouldn't return an error and I'll have to cast the response to bsoncore.Document, then pass it through extractError and if that returns an error I'll process that error using ProcessError. Correct?

Not quite. `ReadWireMessage` will return a wire message but `extractError` expects a command document. You'll have to do this after the wire message goes through the steps to parse it into a `Message`.

> Actually we're checking for non-0 values for poolCleared - See proxy_test:489 on this feature branch.

Ah, my bad. Once we add the functionality to do error handling for well-formed errors, we're need to add tests for that as well.

TomerYakir:
for #2 - I guess we could use failpoints for that as well, right? would a simulated `InterruptedAtShutdown` error be sufficient for this test?

divjotarora:
Yes, that should be sufficient.

TomerYakir:
@divjotarora sorry I forgot to clarify on the expected behavior of ProcessError wrt InterruptedAtShutdown. Is it expected to clear the pool?
Also, does `ProcessError` act on the error code or is it looking into something else?
For reference, configuring the following failpoint:
```
cmd := bson.D{
		{"configureFailPoint", "failCommand"},
		{"mode", "alwaysOn"},
		{"data", bson.D{
			{"failCommands", []string{"update"}},
			{"errorCode", 11660},
		}},
	}
```
Results in:
```
[{ok 0} {errmsg Failing command due to 'failCommand' failpoint} {code 11660} {codeName Location11660}]
```

divjotarora:
The actual error processing functionality mainly looks at error codes and messages. When using `failCommand`, you'll be able to specify the code but not the message. This is sufficient, though, because the message is only used as a fallback if a matching code isn't found. There are some other intricacies regarding `topologyVersion` and pool generation numbers, but these shouldn't affect the testing. They mainly exist to make sure the same server state change does not result in multiple driver-side state changes during a real-life failover/shutdown scenario.

TomerYakir:
Is the error processing of `InterruptedAtShutdown` expected to result in pool clearance?

divjotarora:
Yes, per this [section](https://github.com/mongodb/specifications/blob/83557dbafe8e908eb0559669acfe9f73b561d297/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#not-master-and-node-is-recovering) of the SDAM spec, shutdown error codes (11600 and 91) should always result in a pool clear.

TomerYakir:
@divjotarora RFAL.
I was using the following commit for the driver - fb1610d2e1a64e0bc976c4faa58ce131e39e3d7b.
For some reason error code 11600 didn't result in a pool clear, but 91 did.

proxy.go
```diff
@@ -378,7 +375,16 @@ func (ps *ProxySession) getMongoConnection(rp *readpref.ReadPref) (*MongoConnect
 	if !ok {
 		return nil, fmt.Errorf("bad connection type %T", conn)
 	}
-	return &MongoConnectionWrapper{conn, ec, false, ps.proxy.logger, ps.proxy.config.TraceConnPool}, nil
+	ep, ok := srv.(driver.ErrorProcessor)
+	if !ok {
+		return nil, fmt.Errorf("server ErrorProcessor type assertion failed")
```
divjotarora, line 380: You should never get here, but if you do, `conn` is leaked. I'd recommend moving this to between the `SelectServer` and `Connection()` calls so we don't even attempt connection checkout if this happens.

proxy_test.go
```diff
@@ -441,15 +490,41 @@ func privateTester(t *testing.T, pc ProxyConfig, host string, proxyPort, mongoPo
...
-	enableFailPoint(host, mongoPort, mode)
+	t.Log("*** enable close connection failpoint and run ops")
+	enableFailPointCloseConnection(host, mongoPort, mode)
 	failing = runOps(host, proxyPort, parallelism, true, t, mode, secondaryReads)
 
 	if atomic.LoadInt32(&failing) > 0 {
 		t.Fatalf("ops failures")
 		return
 	}
+	if cleared2 := proxy.GetPoolCleared(); cleared > cleared2 {
```
divjotarora, line 525: Is this supposed to be `cleared2 < cleared`. As it stands, this assertion doesn't account for the `cleared == cleared2` case.

proxy_test.go
```diff
@@ -441,15 +490,41 @@ func privateTester(t *testing.T, pc ProxyConfig, host string, proxyPort, mongoPo
...
+	}
+	cleared := proxy.GetPoolCleared()
+	if cleared == 0 {
+		t.Fatalf("expected pool cleared to be gt 0 but was 0")
+	}
 
-	t.Log("*** enable failpoint and run ops")
-	enableFailPoint(host, mongoPort, mode)
+	t.Log("*** enable close connection failpoint and run ops")
+	enableFailPointCloseConnection(host, mongoPort, mode)
```
divjotarora, line 518: Should we disable the previous fail point? It's possible the server overwrites any existing fail points for a command if a new one is set, in which case the current code works, but we probably don't want to rely on that behavior as fail points are not a well-documented feature.

divjotarora:
@TomerYakir Error code 91 is fine, but I'm not sure why there was no state change when using code 11600. I was able to see a pool clear using the commit hash you mentioned and this script: https://gist.github.com/divjotarora/133a993bbda794c1d2f11df7a391fa4a

TomerYakir:
@divjotarora RFAL. It does work with 11600. I think I may have had a typo at some point which made me think it's not working.

proxy_test.go
```diff
@@ -437,19 +492,50 @@ func privateTester(t *testing.T, pc ProxyConfig, host string, proxyPort, mongoPo
...
+
+	t.Log("*** enable close connection failpoint and run ops")
+	enableFailPointCloseConnection(host, mongoPort, mode)
 	failing = runOps(host, proxyPort, parallelism, true, t, mode, secondaryReads)
 
 	if atomic.LoadInt32(&failing) > 0 {
 		t.Fatalf("ops failures")
 		return
 	}
+	if cleared2 := proxy.GetPoolCleared(); cleared2 < cleared2 {
```
divjotarora, line 536:
```suggestion
	if cleared2 := proxy.GetPoolCleared(); cleared2 < cleared {
```

divjotarora (approved):
Minor comment, but otherwise LGTM.

proxy_test.go
TomerYakir: derp! fixed

PR #60, opened 2020-11-13
Add Evergreen setup

louisaberger:
@TomerYakir -- the mongos int test is still failing, but this should put the structure in place!


server_test.go
```diff
@@ -379,49 +379,51 @@ func checkClient(opts *options.ClientOptions) error {
...
+// 	opts := options.Client().ApplyURI(fmt.Sprintf("mongodb://127.0.0.1:%d", port))
+// 	for i := 0; i < 10; i++ {
+// 		if err := checkClient(opts); err != nil {
+// 			t.Error(err)
+// 		}
+// 	}
+// 	sessCtrCurr := atomic.LoadInt32(&sessCtr)
+
+// 	if sessCtrCurr != int32(20) {
+// 		t.Errorf("expect session counter to be 20 but got %d", sessCtrCurr)
```
TomerYakir, line 420: fwiw this is fixed by changing the counter to be 30 instead of 20. I fixed that on #61.

TomerYakir (changes requested):
Looks like there's a merge conflict. Otherwise LGTM!

server_test.go
louisaberger: Thank you! I'm going to merge in with commented-out, would you mind uncommenting in 61?

TomerYakir: sure!

PR #61, opened 2020-11-16
Connection pool logging flags and integration tests

TomerYakir:
Most of this PR is added logging and integration tests


louisaberger (approved):

util/test_utils.go
```diff
@@ -0,0 +1,95 @@
...
+		hostname, err = os.Hostname()
+		if err != nil {
+			panic(err)
+		}
+	*/
+	hostname = "localhost"
+	return
+}
+
+func GetTestClient(host string, port int, mode MongoConnectionMode, secondaryReads bool, appName string) (*mongo.Client, error) {
```
divjotarora, line 73: nit: Might be useful to call `Connect` in this function so callers get back a `Client` that they can use directly without any additional setup.

util/test_utils.go
```diff
@@ -0,0 +1,95 @@
...
+	var err error
+	mongoPort = 30000
+	proxyPort, err = EphemeralPort()
+	if err != nil {
+		panic(err)
+	}
+	if os.Getenv("MONGO_PORT") != "" {
+		mongoPort, _ = strconv.Atoi(os.Getenv("MONGO_PORT"))
+	}
+	/*
```
divjotarora, line 63: Should this code by removed?

util/test_utils.go
```diff
@@ -0,0 +1,95 @@
...
+		opts.SetAppName(appName)
+	}
+	client, err := mongo.NewClient(opts)
+	if err != nil {
+		return nil, fmt.Errorf("cannot create a mongo client. err: %v", err)
+	}
+	return client, nil
+}
+
+func DisableFailPoint(client *mongo.Client, ctx context.Context) error {
```
divjotarora, line 89: If the function to disable fail points is in this package, should the functions to enable fail points for both error codes and network errors be here as well?

divjotarora (commented):
Mostly small nits.

inttests/proxy_sanity_test.go
```diff
@@ -0,0 +1,328 @@
...
+	"go.mongodb.org/mongo-driver/mongo/options"
+)
+
+const (
+	ParallelClients                = 5
+	InterruptedAtShutdownErrorCode = 11600
+)
+
+func enableFailPointCloseConnection(mongoPort int) error {
+	// cannot fail "find" because it'll prevent certain driver machinery when operating against replica sets to work properly
```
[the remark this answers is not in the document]
TomerYakir: IIRC when I used `find` the server has failed a driver attempt to query `config.sessions`. I can try to reproduce if needed.

util/test_utils.go
TomerYakir: I think I wanted to avoid passing a context to that function so cleanup of `cancelFunc` can happen elsewhere, but I guess I can change the function to accept a context. Agreed?

inttests/proxy_sanity_test.go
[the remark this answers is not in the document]
divjotarora: The driver does not query `config.sessions` internally. It's possible that's an query done from one replica set member to another, but there's been a good amount of work done to ensure those queries do not trigger external failCommand fail points unless the `data.failInternalCommands` option is set. I'm fine leaving this as a fail point on "update", but I'd prefer to remove this line about not being able to fail "find" due to driver machinery because that's not a known issue.

util/test_utils.go
divjotarora: New code that accepts contexts LGTM.

divjotarora (approved):
One small comment remaining, but no need to request another review.

PR #62, opened 2020-11-17
Create request and response timing and errors metric hooks

varunravi98:
This PR adds hooks to measure request and response time and track the number of errors encountered during this process. The hook is designed to be analogous to `ProxyInterceptor` and `ProxyInterceptorFactory`, with `MetricsHook` and `MetricsHookFactory`, respectively. Each `ProxyConfig` now contains a `MetricsHookFactory` field called `CollectorHookFactory`. A non-nil type that implements `CollectorHookFactory` implements the `NewHook()` method, which enables it to create new `MetricsHook`s. Meanwhile, the `ProxySession` type now includes two additional fields: `hooks`, which is a map of `MetricsHook`s, and `isMetricsEnabled`, a bool. When `CreateWorker` is called to generate a new `ProxySession`, `hooks` is populated with `MetricsHook`s generated from the `ProxyConfig`'s `CollectorHookFactory` field (provided that it is non-nil). At the moment, the `MetricsHook`s included are all the ones that are needed for the metrics being tracked, but more can be easily added as necessary. Meanwhile, `isMetricsEnabled` is set to true only if the `CollectorHookFactory` field was non-nil. If it was `nil`, then none of the hooks can be created, and `isMetricsEnabled` is set to false so that any hook updates are ignored. This should only occur during `mongonet`-only tests.

Each type implementing a `MetricHook` implements `StartTimer()`, `StopTimer()`, `SetGauge()`, `AddCounterGauge()`, `IncCounterGauge()`, `DecGauge()`, and `SubGauge()`. The type could wrap (for example) a Prometheus metric, and by implementing these methods, it will be able to perform any basic action on its underlying metric. Any metric hook can be accessed from a `ProxySession` via the hooks map, provided that the programmer uses the same string key to access the metric as they did to insert it into the map in `CreateWorker()`. Lastly, each hook is utilized wherever necessary, using the appropriate method to modify the underlying metric provided that the hook was initialized in the first place (hence all of the `isMetricsEnabled` checks).

### DIFF
--- a/proxy.go
+++ b/proxy.go
@@ -592,7 +592,9 @@ func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper) (*MongoConnect
 		}
 
 		ps.logMessageTrace(ps.proxy.logger, ps.proxy.config.TraceConnPool, resp)
-		responseDurationHook.StopTimer()
+		if ps.isMetricsEnabled {
+			responseDurationHook.StopTimer()
+		}
 
 		// Send message back to user
 		ps.logTrace(ps.proxy.logger, ps.proxy.config.TraceConnPool, "sending back data to user from mongo conn %v", mongoConn.conn.ID())
--- a/proxy.go
+++ b/proxy.go
@@ -785,12 +785,12 @@ func (p *Proxy) CreateWorker(session *Session) (ServerWorker, error) {
 				return nil, err
 			}
 
-			requestErrorsHook, err := ps.proxy.config.CollectorHookFactory.NewHook("processingDuration", "type", "request")
+			requestErrorsHook, err := ps.proxy.config.CollectorHookFactory.NewHook("processingErrors", "type", "request")
 			if err != nil {
 				return nil, err
 			}
 
-			responseErrorsHook, err := ps.proxy.config.CollectorHookFactory.NewHook("processingDuration", "type", "response")
+			responseErrorsHook, err := ps.proxy.config.CollectorHookFactory.NewHook("processingErrors", "type", "response")
 			if err != nil {
 				return nil, err
 			}
--- a/proxy.go
+++ b/proxy.go
@@ -429,25 +429,29 @@ func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper) (*MongoConnect
 	}
 
 	// reading message from client
-	if ps.isMetricsEnabled {
-		err := requestDurationHook.StartTimer()
-		if err != nil {
-			ps.logTrace(ps.proxy.logger, ps.proxy.config.TraceConnPool, "failed to start request duration metric hook timer %v", err)
-		}
-	}
-
 	ps.logTrace(ps.proxy.logger, ps.proxy.config.TraceConnPool, "reading message from client")
 	m, err := ReadMessage(ps.conn)
 	if err != nil {
 		ps.logTrace(ps.proxy.logger, ps.proxy.config.TraceConnPool, "reading message from client fail %v", err)
 		if ps.isMetricsEnabled {
-			requestErrorsHook.IncCounterGauge()
+			hookErr := requestErrorsHook.IncCounterGauge()
+			if hookErr != nil {
+				ps.proxy.logger.Logf(slogger.WARN, "failed to increment requestErrorsHook %v", hookErr)
+			}
 		}
 		if err == io.EOF {
 			return mongoConn, err
 		}
 		return mongoConn, NewStackErrorf("got error reading from client: %v", err)
 	}
+
+	if ps.isMetricsEnabled {
+		hookErr := requestDurationHook.StartTimer()
+		if hookErr != nil {
+			ps.proxy.logger.Logf(slogger.WARN, "failed to start request duration metric hook timer %v", hookErr)
+		}
+	}
+
 	var rp *readpref.ReadPref = ps.proxy.defaultReadPref
 	if ps.proxy.config.ConnectionMode == util.Cluster {
 		// only concerned about OP_MSG at this point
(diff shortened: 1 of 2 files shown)


proxy.go
```diff
@@ -390,11 +406,43 @@ func wrapNetworkError(err error) error {
...
+		}
+		responseErrorsHook, ok = ps.hooks["responseErrorsHook"]
+		if !ok {
+			return nil, fmt.Errorf("could not access the response processing errors metric hook")
+		}
+	}
+
 	// reading message from client
+	if ps.isMetricsEnabled {
+		err := requestDurationHook.StartTimer()
```
TomerYakir, line 433: Let's start the timer after the message is successfully read so we don't count idle client time.
I think we're fine not timing failed attempts to read client messages.

proxy.go
```diff
@@ -390,11 +406,43 @@ func wrapNetworkError(err error) error {
...
+		if err != nil {
+			ps.logTrace(ps.proxy.logger, ps.proxy.config.TraceConnPool, "failed to start request duration metric hook timer %v", err)
```
TomerYakir, line 435: `logTrace` logs at debug-level and is expected to be turned off by default. Let's change to `ps.proxy.logger.Logf(slogger.WARN, ....)`

proxy.go
```diff
@@ -441,15 +501,25 @@ func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper) (*MongoConnect
...
+				requestErrorsHook.IncCounterGauge()
+			}
 			ps.logTrace(ps.proxy.logger, ps.proxy.config.TraceConnPool, "failed to get a new connection. err=%v", err)
 			return nil, NewStackErrorf("cannot get connection to mongo %v using connection mode=%v readpref=%v", err, ps.proxy.config.ConnectionMode, rp)
 		}
 		ps.logTrace(ps.proxy.logger, ps.proxy.config.TraceConnPool, "got new connection %v using connection mode=%v readpref=%v", mongoConn.conn.ID(), ps.proxy.config.ConnectionMode, rp)
 	}
 
+	if ps.isMetricsEnabled {
+		requestDurationHook.StopTimer()
```
TomerYakir, line 514: [q] how do we distinguish between request_proxy_only and request_total?

proxy.go
```diff
@@ -467,13 +537,27 @@ func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper) (*MongoConnect
...
+				responseErrorsHook.IncCounterGauge()
+			}
 			mongoConn.ep.ProcessError(wrapNetworkError(err), mongoConn.conn)
 			return nil, NewStackErrorf("error reading wire message from mongo conn %v: %v", mongoConn.conn.ID(), err)
 		}
+
+		if ps.isMetricsEnabled {
+			err = responseDurationHook.StartTimer()
+			if err != nil {
+				ps.logTrace(ps.proxy.logger, ps.proxy.config.TraceConnPool, "failed to start response duration metric hook timer %v", err)
```
TomerYakir, line 550: same comment about logging

proxy.go
```diff
@@ -482,26 +566,43 @@ func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper) (*MongoConnect
...
+				if ps.isMetricsEnabled {
+					responseErrorsHook.IncCounterGauge()
+				}
 				return nil, NewStackErrorf("error intercepting message %v", err)
 			}
 		}
+
 		ps.logMessageTrace(ps.proxy.logger, ps.proxy.config.TraceConnPool, resp)
+		if ps.isMetricsEnabled {
+			responseDurationHook.StopTimer()
```
TomerYakir, line 596: where do we distinguish between response_total and response_proxy_only?

TomerYakir (changes requested):

TomerYakir:
Small comments from me.
One thing to note is that `mongonet` is actually agnostic to the way metrics are actually implemented, so I'd rephrase `The type must wrap a Prometheus metric` to `The type could (for example) wrap a Prometheus metric`.

proxy.go
```diff
@@ -441,15 +517,31 @@ func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper) (*MongoConnect
...
+				}
+			}
 			ps.logTrace(ps.proxy.logger, ps.proxy.config.TraceConnPool, "failed to get a new connection. err=%v", err)
 			return nil, NewStackErrorf("cannot get connection to mongo %v using connection mode=%v readpref=%v", err, ps.proxy.config.ConnectionMode, rp)
 		}
 		ps.logTrace(ps.proxy.logger, ps.proxy.config.TraceConnPool, "got new connection %v using connection mode=%v readpref=%v", mongoConn.conn.ID(), ps.proxy.config.ConnectionMode, rp)
 	}
 
+	if ps.isMetricsEnabled {
+		requestDurationHook.StopTimer()
```
TomerYakir, line 533: [q] what would happen if we error on line 449 when starting the timer? could that potentially panic?

varunravi98: You're right, that could result in a panic. I could simply set a flag that indicates whether or not `requestDurationHook`'s timer was actually started and check that to stop the timer. The same solution could analogously be applied to `responseDurationHook`'s timer.
Alternatively, I could modify the implementation of `StopTimer()` to return an error if the underlying timer is `nil` (it shouldn't be if `StartTimer()` executed without error, and it will be if an error is returned or if `StartTimer()` wasn't called at all previously). That way, we don't have to add any checks here, but we would enforce that all types implementing `StopTimer()` must check that the underlying timer is non-nil and return an error if it is (or if anything else goes wrong when stopping the timer).

varunravi98: Per offline discussion, the first option was selected. The two flags are `isRequestTimerStarted` and `isResponseTimerStarted`, respectively.

TomerYakir (approved):
LGTM! 👍

PR #64, opened 2020-11-26
Remote Connections API

TomerYakir:
This PR adds the ability to use the proxy over multiple (remote) connections.

The remote connection pool needs to be initialized first by running:
`proxy.AddRemoteConnection(rsName, uri, appName string, trace bool, serverSelectionTimeoutSec, maxPoolSize int, rootCAs *x509.CertPool)`

Then it can be used by passing back the required `rsName` from `InterceptClientToMongo()`. Empty `rsName` indicates the proxy should use the default (local) connection.

Finally, the remote connection should be cleared by `proxy.ClearRemoteConnection(rsName)`.

### DIFF
--- a/inttests/remoteConns.go
+++ b/inttests/remoteConns.go
@@ -1,0 +1,133 @@
+package inttests
+
+import (
+	"context"
+	"fmt"
+	"math/rand"
+	"time"
+
+	"github.com/mongodb/mongonet"
+	"github.com/mongodb/mongonet/util"
+	"github.com/mongodb/slogger/v2/slogger"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+)
+
+const (
+	LocalDbName        = "testlocal"
+	RemoteConnCollName = "test"
+)
+
+func RunProxyConnectionPerformanceRemoteConns(iterations, mongoPort, proxyPort int, hostname string, logger *slogger.Logger, mode util.MongoConnectionMode, goals []ConnectionPerformanceTestGoal,
+	mongoClientFactory util.ClientFactoryFunc,
+	proxyClientFactory util.ClientFactoryFunc,
+) error {
+	for _, goal := range goals {
+		if err := runProxyConnectionPerformanceRemoteConns(iterations, mongoPort, proxyPort, hostname, logger, goal.Workers, goal.AvgLatencyMs, goal.MaxLatencyMs, mode, mongoClientFactory, proxyClientFactory); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func findOne(logger *slogger.Logger, coll *mongo.Collection, goctx context.Context) error {
+	rand.Seed(time.Now().UnixNano())
+	doc := bson.D{}
+	res := coll.FindOne(goctx, bson.D{})
+	if res.Err() != nil {
+		return res.Err()
+	}
+	if err := res.Decode(&doc); err != nil {
+		return err
+	}
+	ix := mongonet.BSONIndexOf(doc, "a")
+	val, _, err := mongonet.GetAsInt(doc[ix])
+	if err != nil {
+		return err
+	}
+	if coll.Database().Name() == util.RemoteDbNameForTests {
+		if val != 2 {
+			return fmt.Errorf("got unexpected value=%v", val)
+		}
+	} else {
+		if val != 1 {
+			return fmt.Errorf("got unexpected value=%v", val)
+		}
+	}
+
+	return nil
+}
+
+func runRemoteConns(logger *slogger.Logger, client *mongo.Client, workerNum int, ctx context.Context) (time.Duration, bool, error) {
+	rand.Seed(time.Now().UnixNano())
+	start := time.Now()
+
+	localColl := client.Database(LocalDbName).Collection(RemoteConnCollName)
+	remoteColl := client.Database(util.RemoteDbNameForTests).Collection(RemoteConnCollName)
+
+	coll := localColl
+	if workerNum%2 == 0 {
+		coll = remoteColl
+	}
+
+	if err := findOne(logger, coll, ctx); err != nil {
+		return 0, false, err
+	}
+
+	elapsed := time.Since(start)
+	logger.Logf(slogger.DEBUG, "worker-%v finished after %v", workerNum, elapsed)
+	return elapsed, true, nil
+}
+
+func runProxyConnectionPerformanceRemoteConns(iterations, mongoPort, proxyPort int, hostname string, logger *slogger.Logger, workers int, targetAvgLatencyMs, targetMaxLatencyMs int64, mode util.MongoConnectionMode,
+	mongoClientFactory util.ClientFactoryFunc,
+	proxyClientFactory util.ClientFactoryFunc,
+) error {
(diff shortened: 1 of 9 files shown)


TomerYakir:
Ping @louisaberger and @divjotarora

proxy.go
```diff
@@ -701,7 +80,46 @@ func NewProxy(pc ProxyConfig) (Proxy, error) {
 	return p, nil
 }
 
-func getMongoClient(p *Proxy, pc ProxyConfig, ctx context.Context) (*mongo.Client, error) {
+func getMongoClientFromUri(p *Proxy, uri, appName string, trace bool, serverSelectionTimeoutSec, maxPoolSize int, rootCAs *x509.CertPool, ctx context.Context) (*mongo.Client, error) {
+	opts := options.Client()
+	opts.ApplyURI(uri).
+		SetAppName(appName).
+		SetPoolMonitor(&event.PoolMonitor{
```
divjotarora, line 87: nit: Instead of duplicating the code to declare a `PoolMonitor` here and in `getMongoClientFromProxyConfig`, would it be possible to have a helper function like `getPoolMonitor()` or maybe even `getBaseClientOptions()` to construct Client options that are the same in both functions?

proxy_session.go
```diff
@@ -0,0 +1,657 @@
...
+			}
+			return mongoConn, nil
+		}
+		if m == nil {
+			// already responded
+			return mongoConn, nil
+		}
+	}
+	if mongoConn == nil || !mongoConn.expirableConn.Alive() {
+		var topology *topology.Topology
```
divjotarora, line 399: nit: You can remove the if/else to reduce complexity a bit:

```
topology := ps.proxy.topology
if remoteRs != "" {
    // logic that's currently in the "else" clause
}
```

divjotarora (approved):
This looks fine to me, but I'm not sure I have all of the context required to thoroughly review this.

proxy.go
```diff
@@ -737,14 +109,36 @@ func getMongoClient(p *Proxy, pc ProxyConfig, ctx context.Context) (*mongo.Clien
...
 	}
-	if pc.ConnectionPoolHeartbeatIntervalMs > 0 {
-		opts.SetHeartbeatInterval(time.Duration(pc.ConnectionPoolHeartbeatIntervalMs) * time.Millisecond)
+	if connectionPoolHeartbeatIntervalMs > 0 {
+		opts.SetHeartbeatInterval(time.Duration(connectionPoolHeartbeatIntervalMs) * time.Millisecond)
 	}
+	return opts
+}
+
+func getMongoClientFromUri(p *Proxy, uri, appName string, trace bool, serverSelectionTimeoutSec, maxPoolSize, maxPoolIdleTimeSec, connectionPoolHeartbeatIntervalMs int, rootCAs *x509.CertPool, ctx context.Context) (*mongo.Client, error) {
```
louisaberger, line 123: I didn't find it super obvious when these two functions should be used and how they're different -- perhaps it's worth specifying in this one that it's for remote connections?

proxy.go
```diff
@@ -763,6 +157,37 @@ func getMongoClient(p *Proxy, pc ProxyConfig, ctx context.Context) (*mongo.Clien
 	return mongo.Connect(ctx, opts)
 }
 
+func (p *Proxy) AddRemoteConnection(rsName, uri, appName string, trace bool, serverSelectionTimeoutSec, maxPoolSize, maxPoolIdleTimeSec, connectionPoolHeartbeatIntervalMs int, rootCAs *x509.CertPool) error {
+	p.logger.Logf(slogger.DEBUG, "adding remote connection for %s", rsName)
+	if _, ok := p.remoteConnections.Load(rsName); ok {
```
louisaberger, line 162: [nit] Rather than "ok", should this be something like "alreadyAdded"?

proxy_session.go
```diff
@@ -0,0 +1,655 @@
+package mongonet
```
louisaberger, line 1: It looks like a lot of this is just moved over from the proxy.go file -- are there changes here we should be looking at, or is this just a refactor?

inttests/remoteConns.go
```diff
@@ -0,0 +1,133 @@
...
+
+func runRemoteConns(logger *slogger.Logger, client *mongo.Client, workerNum int, ctx context.Context) (time.Duration, bool, error) {
+	rand.Seed(time.Now().UnixNano())
+	start := time.Now()
+
+	localColl := client.Database(LocalDbName).Collection(RemoteConnCollName)
+	remoteColl := client.Database(util.RemoteDbNameForTests).Collection(RemoteConnCollName)
+
+	coll := localColl
+	if workerNum%2 == 0 {
```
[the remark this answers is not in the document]
TomerYakir: we'd like to simulate a workload in which 50% of the connections are local and 50% are remote.
I added this as a comment.

proxy.go
TomerYakir: good point - added!

proxy.go
TomerYakir: changed!

proxy_session.go
TomerYakir: Most of it is just a refactor.
The only changes are to the `getMongoConnection` method. and anything that references `remoteRs` in the `doLoop` method

inttests/run_integration_tests_mongos_mode.sh
```diff
@@ -5,21 +5,28 @@ echo ### RUNNING INTEGRATION TESTS
...
 
 $MONGO_DIR/mongo --port $MONGO_PORT --eval "rs.initiate({'_id': 'proxytest', 'protocolVersion': 1, 'writeConcernMajorityJournalDefault': false, 'members': [{_id:0, host:'$HOSTNAME:$MONGO_PORT'},{_id:1, host:'$HOSTNAME:$((MONGO_PORT+1))',priority:0},{_id:2, host:'$HOSTNAME:$((MONGO_PORT+2))',priority:0}]})"
 
-sleep 10 # let replica set reach steady state
+MONGO_PORT=40000
+$MONGO_DIR/mongod --port $MONGO_PORT --dbpath `pwd`/dbpath/4 --logpath `pwd`/dbpath/4/mongod.log --fork --setParameter enableTestCommands=1 --replSet proxytest2
+$MONGO_DIR/mongod --port $((MONGO_PORT+1)) --dbpath `pwd`/dbpath/5 --logpath `pwd`/dbpath/5/mongod.log --fork --setParameter enableTestCommands=1 --replSet proxytest2
+$MONGO_DIR/mongod --port $((MONGO_PORT+2)) --dbpath `pwd`/dbpath/6 --logpath `pwd`/dbpath/6/mongod.log --fork --setParameter enableTestCommands=1 --replSet proxytest2
+
+$MONGO_DIR/mongo --port $MONGO_PORT --eval "rs.initiate({'_id': 'proxytest2', 'protocolVersion': 1, 'writeConcernMajorityJournalDefault': false, 'members': [{_id:0, host:'$HOSTNAME:$MONGO_PORT'},{_id:1, host:'$HOSTNAME:$((MONGO_PORT+1))',priority:0},{_id:2, host:'$HOSTNAME:$((MONGO_PORT+2))',priority:0}]})"
```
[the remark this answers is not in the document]
TomerYakir: this was really a copy-paste from our proxy script. that setting is actually unnecessary.

louisaberger (approved):

PR #65, opened 2020-12-01
API for retrying operation

TomerYakir:
`NewProxyRetryError` to allow signalling the session loop to retry a client message (on a remote RS).
The main intention is that it'll be used by `InterceptMongoToClient`, but it can be potentially used by `InterceptClientToMongo` as well.

### DIFF
--- a/bsonutil.go
+++ b/bsonutil.go
@@ -389,5 +389,5 @@ func BSONGetValueByNestedPathForTests(doc bson.D, nestedPath string, arrIndex in
 			return tempDoc[ix].Value
 		}
 	}
-	return nil
+	return tempDoc
 }
--- a/bsonutil.go
+++ b/bsonutil.go
@@ -376,6 +376,9 @@ func BSONGetValueByNestedPathForTests(doc bson.D, nestedPath string, arrIndex in
 		case bson.D:
 			tempDoc = v
 		case primitive.A:
+			if arrIndex < 0 || len(v) <= arrIndex {
+				return v
+			}
 			switch v2 := v[arrIndex].(type) {
 			case bson.D:
 				tempDoc = v2
--- a/bsonutil.go
+++ b/bsonutil.go
@@ -362,3 +362,29 @@ func BSONWalkHelp(doc bson.D, path []string, visitor BSONWalkVisitor, inArray bo
 
 	return doc, nil
 }
+
+func BSONGetValueByNestedPathForTests(doc bson.D, nestedPath string, arrIndex int) interface{} {
+	tempDoc := doc
+	parts := strings.Split(nestedPath, ".")
+	var ix int
+	for _, p := range parts {
+		ix = BSONIndexOf(tempDoc, p)
+		if ix < 0 {
+			return nil
+		}
+		switch v := tempDoc[ix].Value.(type) {
+		case bson.D:
+			tempDoc = v
+		case primitive.A:
+			switch v2 := v[arrIndex].(type) {
+			case bson.D:
+				tempDoc = v2
+			default:
+				return v2
+			}
+		default:
+			return tempDoc[ix].Value
+		}
+	}
+	return nil
+}
--- a/inttests/int_test_utils.go
+++ b/inttests/int_test_utils.go
@@ -83,6 +83,7 @@ func (srf *SimulateRetryFixer) InterceptMongoToClient(m Message) (Message, error
 		val := BSONGetValueByNestedPathForTests(doc, "cursor.firstBatch.val", 0)
 		if v, ok := val.(int32); ok {
 			fmt.Println("SimulateRetryFixer::got ", v)
+			// trigger a retry error only if the server responds with a particular value
 			if v == util.RetryOnRemoteVal {
 				return mm, NewProxyRetryError(srf.OriginalMessage, util.RemoteRsName)
 			}
--- a/inttests/int_test_utils.go
+++ b/inttests/int_test_utils.go
@@ -52,7 +52,47 @@ func (myf *MyFactory) NewInterceptor(ps *ProxySession) (ProxyInterceptor, error)
 	return &MyInterceptor{ps, myf.mode, myf.mongoPort, myf.proxyPort, myf.disableIsMaster}, nil
 }
 
-type MyResponseInterceptor struct {
+type SimulateRetryFixer struct {
+	OriginalMessage Message
+}
+
+func (srf *SimulateRetryFixer) InterceptMongoToClient(m Message) (Message, error) {
+	switch mm := m.(type) {
+	case *MessageMessage:
+		var err error
+		var bodySection *BodySection = nil
+		for _, section := range mm.Sections {
+			if bs, ok := section.(*BodySection); ok {
+				if bodySection != nil {
+					return mm, NewStackErrorf("OP_MSG should not have more than one body section!  Second body section: %v", bs)
+				}
+				bodySection = bs
+			} else {
+				// MongoDB 3.6 does not support anything other than body sections in replies
+				return mm, NewStackErrorf("OP_MSG replies with sections other than a body section are not supported!")
+			}
+		}
+
+		if bodySection == nil {
+			return mm, NewStackErrorf("OP_MSG should have a body section!")
+		}
+		doc, err := bodySection.Body.ToBSOND()
+		if err != nil {
+			return mm, err
+		}
+		val := BSONGetValueByNestedPathForTests(doc, "cursor.firstBatch.val", 0)
+		if v, ok := val.(int32); ok {
+			if v == util.RetryOnRemoteVal {
+				return mm, NewProxyRetryError(srf.OriginalMessage, util.RemoteRsName)
+			}
+		}
+		return mm, nil
+	default:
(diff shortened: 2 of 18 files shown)


proxy_session.go
```diff
@@ -89,18 +89,42 @@ func (ps *ProxySession) Stats() bson.D {
...
 			}
 			return
 		}
 	}
+	if !shouldRetry {
+		return
+	}
+	// retry the message on another rs in case of a ProxyRetryError
+	for {
+		ps.mongoConn, err = ps.doLoop(ps.mongoConn, retryError.MsgToRetry, retryError.RetryOnRs, true)
```
divjotarora, line 116: Is it possible to get into a situation where we're retrying but `ps.mongoConn` is still valid? If that were to happen, the code in `doLoop` would recognize that it's retrying but would not check out a new connection for the remote RS because the existing one is non-nil and is alive.

proxy_session.go
```diff
@@ -282,7 +306,7 @@ func wrapNetworkError(err error) error {
 	return driver.Error{Message: err.Error(), Labels: labels, Wrapped: err}
 }
 
-func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper) (*MongoConnectionWrapper, error) {
+func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper, m Message, retryOnRs string, isRetry bool) (*MongoConnectionWrapper, error) {
```
divjotarora, line 309: nit: From an API perspective, it may be cleaner to have this be
```
doLoop(mongoConn *MongoConnectionWrapper, retryErr *ProxyRetryError) (*MongoConnectionWrapper, error)
```

The `isRetry` bool would be replaced by `retryErr != nil` checks.

divjotarora (commented):
I don't have a lot of context for this PR, but the changes look fine overall. I just had one question about connection state and how it relates to retries.

proxy_session.go
TomerYakir: Not really - we would get to the "retry" loop only in case the first loop returns with `ProxyRetryError`, in which case the `ps.mongoConn` is guaranteed to be closed (see `proxy_session:98` on that feature branch)

divjotarora: Oh I see. The previous `doLoop` call would've returned `(nil, pre)` so the next call for the retry would be forced to enter the connection checkout code.

TomerYakir:
@louisaberger comments by Divjot were addressed. RFAL!

proxy_session.go
```diff
@@ -89,18 +89,42 @@ func (ps *ProxySession) Stats() bson.D {
...
-		ps.mongoConn, err = ps.doLoop(ps.mongoConn)
+		ps.mongoConn, err = ps.doLoop(ps.mongoConn, nil)
 		if err != nil {
 			if ps.mongoConn != nil {
 				ps.mongoConn.Close(ps)
 			}
+			retryError, shouldRetry = err.(*ProxyRetryError)
+			if shouldRetry {
+				ps.logger.Logf(slogger.WARN, "%v", retryError)
+				break
```
louisaberger, line 103: This is a bit more confusing to parse the control flow here -- do you think it would be simpler to just call out to a different function here (or do the retry here) and then return? Rather than breaking, and then having a case at the end?

proxy_session.go
```diff
@@ -282,7 +306,7 @@ func wrapNetworkError(err error) error {
 	return driver.Error{Message: err.Error(), Labels: labels, Wrapped: err}
 }
 
-func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper) (*MongoConnectionWrapper, error) {
+func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper, pre *ProxyRetryError) (*MongoConnectionWrapper, error) {
```
louisaberger, line 309: Ah, I interpreted `pre` as previous below :) Maybe it would be clearer to just call it `retryErrror` or something?

proxy_session.go
TomerYakir: done

bsonutil.go
```diff
@@ -362,3 +362,32 @@ func BSONWalkHelp(doc bson.D, path []string, visitor BSONWalkVisitor, inArray bo
...
+	parts := strings.Split(nestedPath, ".")
+	var ix int
+	for _, p := range parts {
+		ix = BSONIndexOf(tempDoc, p)
+		if ix < 0 {
+			return nil
+		}
+		switch v := tempDoc[ix].Value.(type) {
+		case bson.D:
+			tempDoc = v
```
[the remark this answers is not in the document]
TomerYakir: good catch! fixed.

inttests/int_test_utils.go
```diff
@@ -250,36 +292,45 @@ func (myi *MyInterceptor) InterceptClientToMongo(m Message) (Message, ResponseIn
...
-		if !myi.disableStreamingIsMaster {
+		switch strings.ToLower(doc[0].Key) {
+		case "ismaster":
+			// streaming isMaster is enabled. no need to fix
+			if !myi.disableStreamingIsMaster {
+				return mm, nil, rsName, nil
+			}
+			// fixing isMaster request when streamingIsMaster is disabled
+			if idx := BSONIndexOf(doc, "maxAwaitTimeMS"); idx >= 0 {
+				doc = append(doc[:idx], doc[idx+1:]...)
```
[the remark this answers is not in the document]
TomerYakir: if `BSONIndexOf` returns a >0 value, this should never panic (I double checked that with various `idx` values).

proxy_session.go
```diff
@@ -86,15 +86,41 @@ func (ps *ProxySession) Stats() bson.D {
 	}
 }
 
+func (ps *ProxySession) doRetryLoop(retryError *ProxyRetryError) {
+	// retry the message on another rs in case of a ProxyRetryError. This function assumes that ps.mongoConn is closed or nil.
+	var err error
+	for {
```
divjotarora, line 92: Is there a path that would cause this loop to execute more than once? Seems like the execution paths for both nil and non-nil errors from `doLoop` end in an explicit `return` statement, so the loop will terminate after a single iteration.

TomerYakir: good catch! working on a fix now (will ping you when done).

TomerYakir: @divjotarora should be fine now.

proxy_session.go
```diff
@@ -86,15 +86,42 @@ func (ps *ProxySession) Stats() bson.D {
 	}
 }
 
+func (ps *ProxySession) doRetryLoop(retryError *ProxyRetryError) {
+	// retry the message on another rs in case of a ProxyRetryError. This function assumes that ps.mongoConn is closed or nil.
+	var err error
+	retryOnRs := retryError.RetryOnRs
+	for {
+		ps.mongoConn, err = ps.doLoop(ps.mongoConn, retryError, retryOnRs)
```
divjotarora, line 94: After the first iteration, we might be passing a nil `retryError` but a non-empty `retryOnRs`. Is this intentional?

inttests/int_test_utils.go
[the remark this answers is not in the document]
louisaberger: So what if the doc length is 2, and idx=1? Wouldn't doc[idx+1] panic in that case?

proxy_session.go
```diff
@@ -89,18 +89,42 @@ func (ps *ProxySession) Stats() bson.D {
...
+	}
+	// retry the message on another rs in case of a ProxyRetryError. Note that ps.mongoConn is guaranteed to be closed/nil by the previous loop.
+	for {
+		ps.mongoConn, err = ps.doLoop(ps.mongoConn, retryError)
+		if err != nil {
+			if ps.mongoConn != nil {
+				ps.mongoConn.Close(ps)
+			}
+			if err != io.EOF {
+				ps.logger.Logf(slogger.WARN, "error doing loop: %v", err)
```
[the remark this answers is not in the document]
louisaberger: not sure if you saw this one?

proxy_session.go
TomerYakir: Yes - the first iteration gets the client message from `retryError`, but subsequent iterations should get it from the client.

proxy_session.go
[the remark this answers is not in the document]
TomerYakir: I did and fixed that. Not sure why github doesn't mark this as outdated...
https://github.com/mongodb/mongonet/pull/65/commits/e5ab86d5349763f4535e380a10df9366c6781ee0

inttests/int_test_utils.go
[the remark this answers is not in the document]
TomerYakir: `doc[idx+1]` would panic, but `doc[idx+1:]` wouldn't. I double-checked using this snippet:
```
doc := []int{1, 2}
idx := 1
// fmt.Println(doc[2]) panics, but...
// fmt.Println(doc[2:]) return an empty slice
doc = append(doc[:idx], doc[idx+1:]...)
fmt.Println(doc) // prints 1
```

divjotarora (approved):

louisaberger (approved):

harish-chandramowli (approved):

PR #69, opened 2021-01-27
Cursor server association support

tolsen:
This adds support to mongonet for helping clients to associate a cursor with a particular server.

Louisa, I am assigning to you for now because Tomer is out until Sunday.  If we don't need this merged until next week then we can add Tomer as a reviewer.

### DIFF
--- a/inttests/int_test_utils.go
+++ b/inttests/int_test_utils.go
@@ -13,6 +13,7 @@ import (
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/bson/primitive"
 	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/x/mongo/driver/address"
 )
 
 const (
@@ -56,7 +57,7 @@ type SimulateRetryFixer struct {
 	OriginalMessage Message
 }
 
-func (srf *SimulateRetryFixer) InterceptMongoToClient(m Message) (Message, error) {
+func (srf *SimulateRetryFixer) InterceptMongoToClient(m Message, _ address.Address) (Message, error) {
 	switch mm := m.(type) {
 	case *MessageMessage:
 		var err error
@@ -158,7 +159,7 @@ func fixIsMasterDirect(doc bson.D, mongoPort, proxyPort int) (SimpleBSON, error)
 	return SimpleBSONConvert(doc)
 }
 
-func (mri *IsMasterFixer) InterceptMongoToClient(m Message) (Message, error) {
+func (mri *IsMasterFixer) InterceptMongoToClient(m Message, _ address.Address) (Message, error) {
 	switch mm := m.(type) {
 	case *ReplyMessage:
 		var err error
--- a/inttests/int_test_utils.go
+++ b/inttests/int_test_utils.go
@@ -288,7 +288,7 @@ func (myi *MyInterceptor) InterceptClientToMongo(m Message) (
 			panic(err)
 		}
 		mm.Query = qb
-		return mm, &IsMasterFixer{myi.mode, myi.mongoPort, myi.proxyPort}, "", nil
+		return mm, &IsMasterFixer{myi.mode, myi.mongoPort, myi.proxyPort}, "", "", nil
 	case *MessageMessage:
 		var err error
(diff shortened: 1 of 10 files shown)


wire.go
```diff
@@ -145,5 +156,16 @@ type MessageMessage struct {
...
+func (mm *MessageMessage) BodyDoc() bsoncore.Document {
+	// when parsed in in parseMessageMessage we checked that there was exactly one BodySection
+	// as such we can stop as soon as we find it
+	for _, sec := range mm.Sections {
+		if bodySection, ok := sec.(*BodySection); ok && bodySection != nil {
+			return bsoncore.Document(bodySection.Body.BSON)
+		}
+	}
+
+	panic(fmt.Errorf("unreachable"))
```
TomerYakir, line 170: could we rephrase as something like `no body section found for OP_MSG`?

session.go
```diff
@@ -174,10 +172,12 @@ func (s *Session) RespondToCommand(clientMessage Message, doc SimpleBSON) error
 					doc,
 				},
 			},
-			bsoncore.Document(doc.BSON),
 		}
 		return SendMessage(rm, s.conn)
 
+	case OP_GET_MORE:
+		return errors.New("Internal error")
```
TomerYakir, line 179: [q] do we want to be more specific about the error here?

TomerYakir (changes requested):
Added a few minor comments.
Do you suppose we can add an integration test to test out pinned server selection on mongonet?

wire.go
tolsen: I believe we would have errored when we constructed the MessageMessage in the first place.  That's why I treated this more like an internal contract violation.  I'll see how onerous it is to thread an error response all the way up

tolsen: it's not that bad. i'll return an error

session.go
tolsen: fixed!

tolsen:
@TomerYakir comments addressed!

tolsen:
> Added a few minor comments.
> Do you suppose we can add an integration test to test out pinned server selection on mongonet?

just saw this question now.  i'll look to see what can be easily added

TomerYakir:
@louisaberger I added an int test for this and confirmed it's failing when the cursor isn't associated with the server address.
ready for review!

inttests/int_test_utils.go
```diff
@@ -31,6 +32,7 @@ func insertDummyDocs(client *mongo.Client, numOfDocs int, ctx context.Context) e
 	if _, err := coll.InsertMany(ctx, docs); err != nil {
 		return fmt.Errorf("initial insert failed. err: %v", err)
 	}
+	fmt.Println("inserted", numOfDocs, "dummy docs")
```
louisaberger, line 35: Do we still need this statement?

session.go
```diff
@@ -125,6 +124,7 @@ func (s *Session) RespondToCommandMakeBSON(clientMessage Message, args ...interf
 	return s.RespondToCommand(clientMessage, doc2)
 }
 
+// do not call with OP_GET_MORE
```
louisaberger, line 127: Could you add a comment about _why_ we shouldn't call with OP_GET_MORE?

wire.go
```diff
@@ -49,8 +53,15 @@ type ReplyMessage struct {
 	StartingFrom   int32
 	NumberReturned int32
 
-	Docs       []SimpleBSON
-	CommandDoc bsoncore.Document
+	Docs []SimpleBSON
+}
```
louisaberger, line 57: Why do we now calculate this each time in the new function instead?

inttests/int_test_utils.go
TomerYakir: removed!

session.go
tolsen: that function doesn't support OP_GET_MORE

wire.go
tolsen: it's a derived field.  I don't see how the below function CommandDoc() is even remotely expensive to call.

The idea is less state, fewer bugs.  This way I got to remove a bunch of code that was intended to keep this field up to date.  I found myself having to add more of that code.  It just seems really error prone.

louisaberger (approved):

PR #73, opened 2021-03-24
Fix pinned server selector for remote topologies

TomerYakir:


TomerYakir:
this fix isn't necessary

PR #76, opened 2021-06-01
Support Proxy Protocol

TomerYakir:
This PR implements the proxy protocol so that we can retain the original client IP on connections that go through another layer of load balancer/proxy.

It also adds SSL tests for the proxy server.

Tested locally that TestCommonProxyProtocolInt passes.

### DIFF
--- a/inttests/pinnedServer_test.go
+++ b/inttests/pinnedServer_test.go
@@ -14,7 +14,7 @@ import (
 	"go.mongodb.org/mongo-driver/mongo/readpref"
 )
 
-func TestPinnedServer(t *testing.T) {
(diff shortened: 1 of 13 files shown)


inttests/pinnedServer_test.go
```diff
@@ -14,7 +14,7 @@ import (
 	"go.mongodb.org/mongo-driver/mongo/readpref"
 )
 
-func TestPinnedServer(t *testing.T) {
+func TestCommonPinnedServer(t *testing.T) {
```
TomerYakir, line 17: drive-by fix - we weren't running this test

tolsen (approved):

PR #84, opened 2021-06-24
Update ResponseInterceptor Interface to ProcessExecutionStats

jr22:
Update the ResponseInterceptor interface to include ExtractExecutionStats. An initial timestamp will be taken once a This should account for the following edge cases: 
- responseInterceptor is `nil`
- responseInterceptor is not used since proxy returns message itself (CMD_I_HANDLE)

Evergreen patch for mongonet: 
https://evergreen.mongodb.com/version/60d5f2f90ae60671e32d0f5d

Evergreen patch for proxy will fail because of update to the ResponseInterceptor interface. Will put up a PR shortly in the proxy repo with changes that allow those to pass

### DIFF
--- a/proxy_session.go
+++ b/proxy_session.go
@@ -47,6 +47,7 @@ type MetricsHookFactory interface {
 
 type ResponseInterceptor interface {
 	InterceptMongoToClient(m Message, serverAddress address.Address, isRemote bool) (Message, error)
+	ExtractExecutionTime(startTime time.Time, pausedExecutionTime int64)
 }
 
 type ProxyInterceptor interface {
@@ -413,9 +414,15 @@ func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper, retryError *Pr
 	ps.logMessageTrace(ps.proxy.logger, ps.proxy.Config.TraceConnPool, m)
 	var respInter ResponseInterceptor
 	var pinnedAddress address.Address
+	pausedExecutionTime := int64(0)
 	if ps.interceptor != nil {
 		ps.interceptor.TrackRequest(m.Header())
 		m, respInter, remoteRs, pinnedAddress, err = ps.interceptor.InterceptClientToMongo(m, previousRes)
+		defer func () {
+			if respInter != nil {
+				respInter.ExtractExecutionTime(startServerSelection, pausedExecutionTime)
+			}
+		}()
 		if err != nil {
 			if m == nil {
 				if ps.isMetricsEnabled {
@@ -636,6 +643,7 @@ func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper, retryError *Pr
 
 		// Send message back to user
 		ps.logTrace(ps.proxy.logger, ps.proxy.Config.TraceConnPool, "sending back data to user from mongo conn id=%v remoteRs=%s", mongoConn.conn.ID(), remoteRs)
+		startPausedTimer := time.Now()
 		err = SendMessage(resp, ps.conn)
 		if err != nil {
 			if ps.isMetricsEnabled {
@@ -676,6 +684,7 @@ func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper, retryError *Pr
 			}
 			return nil, NewStackErrorf("bad response type from server %T", r)
 		}
+		pausedExecutionTime += time.Now().Sub(startPausedTimer).Microseconds()
 	}
 }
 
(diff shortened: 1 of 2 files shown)


proxy_session.go
```diff
@@ -47,6 +47,7 @@ type MetricsHookFactory interface {
 
 type ResponseInterceptor interface {
 	InterceptMongoToClient(m Message, serverAddress address.Address, isRemote bool) (Message, error)
+	ExtractExecutionTime(startTime time.Time, pausedExecutionTime int64)
```
jr22, line 50: went back and forth on whether this should return an error. i decided **not** to return an error for the following reason: 
* in the proxy code, if we go to set latency stats with an invalid readwritetype i believe there should be an error. however, i don't think this error needs to be surfaced to mongonet. We should not return an error to the user in this situation given it would be an internal, implementation error. Instead, i think the error should be logged within the proxy? 

let me know what you think!

TomerYakir: agreed - we generally do not return an error to the user on non-blocking internal implementation failures

jr22:
@TomerYakir i can't seem to add any reviewers to this PR due to permissions issues. Do you mind being a solo reviewer or I can mention on my other related PR to check this one out as well?

proxy_session.go
TomerYakir: reviewing the name again - could you add a comment clarifying what's expected from the implementation?

proxy_session.go
```diff
@@ -413,9 +414,15 @@ func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper, retryError *Pr
 	ps.logMessageTrace(ps.proxy.logger, ps.proxy.Config.TraceConnPool, m)
 	var respInter ResponseInterceptor
 	var pinnedAddress address.Address
+	pausedExecutionTime := int64(0)
```
TomerYakir, line 417: [nit] should we call this `pausedExecutionTimeMicros`?

proxy_session.go
```diff
@@ -676,6 +684,7 @@ func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper, retryError *Pr
 			}
 			return nil, NewStackErrorf("bad response type from server %T", r)
 		}
+		pausedExecutionTime += time.Now().Sub(startPausedTimer).Microseconds()
```
TomerYakir, line 687: [q] there are quite a few `return` statements above this line (for example - `if !r.HasMoreToCome()`. Should we have a 0 `pausedExecutionTime` in those cases?

TomerYakir:
> @TomerYakir i can't seem to add any reviewers to this PR due to permissions issues. Do you mind being a solo reviewer or I can mention on my other related PR to check this one out as well?

I think they'll need to request permissions for this repo via MANA so that you can add them as reviewers.

proxy_session.go
jr22: yeah, i am struggling to think of the most descriptive name for this function. adding a comment here to clarify but open to other naming suggestions too

proxy_session.go
jr22: great catch. I moved this call up to right below the call to `sendMessage`. i was trying to avoid including the time after we already sent the message in the execution time, but I think incrementing `pausedExecutionTime` after every possible return got pretty messy. Seems like the time after `sendMessage` will be pretty negligible anyway. let me know what you think!

Salil999: I'd lean in favor of making it a bit more messy and having it to be more accurate. I am not too familiar with all of the logic past `sendMessage` but if anything in there is expensive then that might contribute to discrepancies over time. But I think Tomer can give a better answer on it.

proxy_session.go
```diff
@@ -47,6 +47,9 @@ type MetricsHookFactory interface {
 
 type ResponseInterceptor interface {
 	InterceptMongoToClient(m Message, serverAddress address.Address, isRemote bool) (Message, error)
+	// ExtractExecutionTime records the execution time of an operation from startTime, subtracting
+	// time while execution was paused, pausedExecutionTimeMicros (i.e. while sending the message back to client)
+	ExtractExecutionTime(startTime time.Time, pausedExecutionTimeMicros int64)
```
TomerYakir, line 52: [q] how about renaming this to `ProcessExecutionTime`? I feel this better represents the implementation-agnostic nature of this interface method

proxy_session.go
TomerYakir: I like the comment, but added a renaming suggestion

proxy_session.go
TomerYakir: the current implementation lgtm!

TomerYakir (approved):

Salil999 (approved):

PR #87, opened 2021-07-14
Support test profiling

tolsen:
Evergreen: https://evergreen.mongodb.com/version/60ef1e66c9ec44252981bc05


inttests/blogapp.go
```diff
@@ -124,7 +126,10 @@ func runBlogApp(logger *slogger.Logger, client *mongo.Client, workerNum int, ctx
 	}
 
 	elapsed := time.Since(start)
-	logger.Logf(slogger.DEBUG, "worker-%v finished after %v", workerNum, elapsed)
+
+	funcName := runtime.FuncForPC(reflect.ValueOf(funcs[ix]).Pointer()).Name()
+
+	logger.Logf(slogger.INFO, "worker-%v called %v and finished after %v", workerNum, funcName, elapsed)
```
TomerYakir, line 132: [q] do we need to apply this on the other benchmarks (`findone` for example) as well?

inttests/testrunner.go
```diff
@@ -170,6 +174,14 @@ func RunIntTest(mode util.MongoConnectionMode, maxPoolSize, workers int, targetA
 	) error,
 	blockCommands map[string]time.Duration,
 ) {
+	callerName := getCallerName()
+
+	runtime.SetBlockProfileRate(100)
+	runtime.SetMutexProfileFraction(5)
+
+	defer writeOutProfile(callerName, "block", 0)
+	defer writeOutProfile(callerName, "mutex", 0)
```
TomerYakir, line 183: [q] are we interested in other profiles as well?

inttests/blogapp.go
tolsen: good idea

inttests/testrunner.go
tolsen: possibly.  I'll look

inttests/blogapp.go
tolsen: @TomerYakir I'm confused.  Is the findone you're talking about blogPostFindOne?  if so, that function is one of the funcs called by this function.

TomerYakir: no, I'm referring to `findone.go:runFind`

tolsen: but runFind doesn't call a random sub-benchmark like blogapp does.  I believe the change here is just to log the name of the sub-benchmark that I've called.  A search for rand.Intn throughout the code shows there's nothing else like this in the tests.

inttests/testrunner.go
tolsen: how about *all* the profiles? lol

tolsen:
new evergreen: https://evergreen.mongodb.com/version/60f0813c61837d21fc937888?redirect_spruce_users=true

tolsen:
@TomerYakir let me know what you think of the changes and if you'd like anything else.

inttests/blogapp.go
TomerYakir: ah got it - makes sense!

inttests/testrunner.go
```diff
@@ -203,6 +214,43 @@ func RunIntTest(mode util.MongoConnectionMode, maxPoolSize, workers int, targetA
...
+	if callerFunc == nil {
+		panic("could not get func for caller PC")
+	}
+	fullCallerName := callerFunc.Name()
+	periodIndex := strings.LastIndex(fullCallerName, ".")
+	return fullCallerName[periodIndex+1:]
+}
+
+func writeOutAllProfiles(filenamePrefix string) {
+	for _, profile := range pprof.Profiles() {
```
TomerYakir, line 232: [q/ignorable] do we want to include a CPU profile as well? can be in a separate PR

TomerYakir (approved):
LGTM! one ignorable question

inttests/testrunner.go
tolsen: good catch!  i'd say let's leave that for a separate PR as it is a different API

PR #89, opened 2021-07-14
Extend retry API to allow multiple retries before return

ahmedssulaiman:
We extend the ProxyRetryError API to allow the caller to specify a count for the number of times we can attempt to retry the command. We also extend InterceptClientToMongo to accept a bool (retryFailed) which allows mongonet to specify to the proxy that the last command was a result of a failed retry (so we can send the error back to the user rather than retrying again).


inttests/int_test_utils.go
```diff
@@ -81,6 +81,12 @@ func (ff *FindFixer) InterceptMongoToClient(m Message, address address.Address,
 				// trigger a retry error only if the server responds with a particular value
 				if v == util.RetryOnRemoteVal {
 					return mm, NewProxyRetryError(ff.OriginalMessage, SimpleBSON{}, util.RemoteRsName)
+				} else if v == util.RetryOnRemoteValMultiple && !retryFailed {
+					// Retry on local, should still get RetryOnRemoteValMultiple
+					return mm, NewProxyRetryErrorWithRetryCount(ff.OriginalMessage, SimpleBSON{}, "", 2)
```
TomerYakir, line 86: [q] is there a way we can test how many times the command has retried?

ahmedssulaiman: I'll re-write these tests with configurefailpoint, should be a better way to test this functionality

TomerYakir: As mentioned on Slack, it's ok to defer this to the application tests and use fail points.

TomerYakir (approved):

inttests/int_test_utils.go
ahmedssulaiman: I decided to re-write the tests with configurefailpoint (as I was a little skeptical they worked after re-reading it), and I found a bug in the process. Fixed and new integration tests should verify the retry behavior based on the number of times as well.

ahmedssulaiman:
Should be ready for review. Some implementation notes
1) I cleaned up the retryloop and merged it into the existing loop, allowed for cleaner processing of retryErrors that had multiple retries.
2) Had to make a deep copy of the original message in case of retries. We could let the proxy know that the incomming message is a retry and partially perform its duties based on that (e.g. not prepend the db name) but that will get messy really quickly. Instead, we pass the original message back to the proxy so it can perform any business logic as it needs without being aware of the retry functionality.

proxy_session.go
```diff
@@ -416,6 +399,14 @@ func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper, retryError *Pr
 	ps.logMessageTrace(ps.proxy.logger, ps.proxy.Config.TraceConnPool, m)
 	var respInter ResponseInterceptor
 	var pinnedAddress address.Address
+
+	// Make deep copy of message to retry before Proxy modifies message
+	// in case of retry
+	messageBeforeIntercept, err := ReadMessageFromBytes(m.Serialize())
```
TomerYakir, line 405: [q] could you check whether this doesn't double the effort made on our application? specifically line 1005 on session_proxy.go on the other repo

TomerYakir: I'm also curious looking into any impact on performance and whether we can reuse previous serialization/deserialization

proxy_session.go
```diff
@@ -95,49 +95,32 @@ func (ps *ProxySession) Stats() bson.D {
...
 		if err != nil {
 			if ps.mongoConn != nil {
 				ps.mongoConn.Close(ps)
 			}
 			retryError, shouldRetry = err.(*ProxyRetryError)
 			if shouldRetry {
 				ps.logger.Logf(slogger.WARN, "%v", retryError)
-				ps.doRetryLoop(retryError)
-				break
+				continue
```
TomerYakir, line 116: nice! ✂️

proxy_session.go
ahmedssulaiman: The biggest issue I came across was that the $db portion of the message wasn't really being persisted. e.g. 

```
(After interceptClientToMongo)
retryError.MsgToRetry: [{find test2} {readConcern [{level snapshot}]} {lsid...} {$db eliot1_test}]
messageBeforeIntercept : [{find test2} {readConcern [{level snapshot}]} {lsid ...} {$db test}]
```
where messageBeforeIntercept is the deep copy before intercept.

I just realized 2 things:

1. When we assign `(if retryError != nil) var m Message = retryError.msgToRetry` <-- this needs to be a deep copy
2. We only need to do this deep copy if retryError != nil, so that should minimize any performance impact due to this change.

Will make change according to the above 2 points

proxy_session.go
```diff
@@ -622,8 +618,21 @@ func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper, retryError *Pr
...
+		if retryError != nil && errCode != 0 {
+			if retryError.RetryCount > 1 {
+				// Retry failed, decrement retryCount and retry
+				retryError.RetryCount -= 1
+				return nil, retryError
+			} else {
+				// We use this flag to signify that a retry failed after
+				// the proxy retried a particular command (RetryCount number
+				// of times.)
+				retryFailed = true
```
TomerYakir, line 631: [q] could we rename to something like `retryAttemptsExhausted`? I feel this would otherwise be a bit ambiguous

inttests/int_test_utils.go
```diff
@@ -79,7 +91,9 @@ func (ff *FindFixer) InterceptMongoToClient(m Message, address address.Address,
 			val := BSONGetValueByNestedPathForTests(doc, "cursor.firstBatch.val", 0)
 			if v, ok := val.(int32); ok {
 				// trigger a retry error only if the server responds with a particular value
-				if v == util.RetryOnRemoteVal {
+				if v == util.RetryOnRemoteVal && !retryFailed {
+					// Retried once and succeeded, retryFailed -> false
+					ff.ps.Logf(slogger.ERROR, "util.RetryOnRemoteVal && NOT retryFailed")
```
TomerYakir, line 96: I had to read this a couple of times to figure out why it would be an issue and then I remembered that this block refers to the existing int test.
A couple of suggestions/questions:
1. Could we add some comments to clarify which parts of FindFixer refer to which test? alternatively we may want to have two variants of the fixer - one for the old test and one for the new one
2. How would we know that we got unexpected behaviour other than looking at the logs? should we panic or error using a non-retryable error instead?

inttests/int_test_utils.go
```diff
@@ -91,10 +87,13 @@ func (ff *FindFixer) InterceptMongoToClient(m Message, address address.Address,
...
-				if v == util.RetryOnRemoteVal && !retryFailed {
-					// Retried once and succeeded, retryFailed -> false
-					ff.ps.Logf(slogger.ERROR, "util.RetryOnRemoteVal && NOT retryFailed")
+				if v == util.RetryOnRemoteVal && !retryAttemptsExhausted {
+					// Retried once and succeeded, retryAttemptsExhausted -> false
+					ff.ps.Logf(slogger.ERROR, "util.RetryOnRemoteVal && NOT retryAttemptsExhausted")
 					return mm, NewProxyRetryError(ff.OriginalMessage, SimpleBSON{}, util.RemoteRsName)
+				} else if v == util.RetryOnRemoteVal && retryAttemptsExhausted {
+					// We should not come back here after first retry.
+					panic("retryAttemptsExhausted should not be true after one succesful retry")
```
TomerYakir, line 96: [sp] succesful -> successful

inttests/int_test_utils.go
```diff
@@ -91,10 +87,13 @@ func (ff *FindFixer) InterceptMongoToClient(m Message, address address.Address,
 			val := BSONGetValueByNestedPathForTests(doc, "cursor.firstBatch.val", 0)
 			if v, ok := val.(int32); ok {
 				// trigger a retry error only if the server responds with a particular value
-				if v == util.RetryOnRemoteVal && !retryFailed {
-					// Retried once and succeeded, retryFailed -> false
-					ff.ps.Logf(slogger.ERROR, "util.RetryOnRemoteVal && NOT retryFailed")
+				if v == util.RetryOnRemoteVal && !retryAttemptsExhausted {
+					// Retried once and succeeded, retryAttemptsExhausted -> false
+					ff.ps.Logf(slogger.ERROR, "util.RetryOnRemoteVal && NOT retryAttemptsExhausted")
```
TomerYakir, line 92: [q] is that expected? if so, should we be logging as `DEBUG` rather than `ERROR`?

ahmedssulaiman: Yep - it's expected. I'll just remove this log line altogether (I initially included it for me to debug).

TomerYakir (approved):

PR #90, opened 2021-07-19
Code review comments from proxy

ahmedssulaiman:
To address code review comments from proxy


util.go
```diff
@@ -21,6 +21,9 @@ type ProxyRetryError struct {
 }
 
 func (e *ProxyRetryError) Error() string {
+	if e.RetryOnRs == "" {
+		return fmt.Sprintf("ProxyRetryError - going to retry on local MTM, %v times remaining", e.RetryCount)
```
TomerYakir, line 25: on second thought, let's call it `local replica set` instead of MTM

TomerYakir (approved):
LGTM - please give the commit a meaningful name (like "rephrase proxyRetryError for local retries")

PR #94, opened 2021-09-02
CLOUDP-98421 Add more detailed latency metrics to mongonet

pstefek-mongo:
Jira: https://jira.mongodb.org/browse/CLOUDP-98421 also covers https://jira.mongodb.org/browse/CLOUDP-98878
EVG: https://spruce.mongodb.com/version/6130387ce3c3317c699eda71/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC
Description: This pr allows to capture both db round trip time and break up all latency measures in mongonet by an operationType category which is provided by the request interceptor.

### DIFF
--- a/proxy_session.go
+++ b/proxy_session.go
@@ -558,7 +558,7 @@ func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper, retryError *Pr
 			return nil, NewStackErrorf("error reading wire message from mongo conn id=%v remoteRs=%s. err=%v", mongoConn.conn.ID(), remoteRs, err)
 		}
 
-		respomseDurationTimerStart := time.Now()
+		responseDurationTimerStart := time.Now()
 		if ps.isMetricsEnabled {
 			if !hasReceivedFirstResponse {
 				elaspedDbRoundTripTime := time.Since(dbRoundTripTimerStart).Seconds()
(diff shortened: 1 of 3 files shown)


proxy_session.go
```diff
@@ -34,6 +34,7 @@ type ProxySession struct {
 type MetricsHook interface {
 	StartTimer() error
 	StopTimer()
+	ObserveWithLabels(val float64, labels map[string]string) error
```
pstefek-mongo, line 37: Imho the metrics hook system should let us assign labels on call and not force them at setup. This is what `ObserveWithLabels` is for but I wonder if we could replace the restriction on the other methods. This is just my opinion tho

proxy_session.go
```diff
@@ -556,13 +558,12 @@ func (ps *ProxySession) doLoop(mongoConn *MongoConnectionWrapper, retryError *Pr
...
 
-		isResponseTimerStarted := false
+		respomseDurationTimerStart := time.Now()
 		if ps.isMetricsEnabled {
-			hookErr := responseDurationHook.StartTimer()
-			if hookErr != nil {
-				ps.proxy.logger.Logf(slogger.WARN, "failed to start reponse duration metric hook timer %v", hookErr)
-			} else {
-				isResponseTimerStarted = true
+			if !hasReceivedFirstResponse {
```
pstefek-mongo, line 563: I choose to measure round trip time as time until first response message from the db. Lmk if you think that's a good measure

proxy_session.go
TomerYakir: I agree!

proxy_session.go
TomerYakir: Could you please chat with Julia Ruddy on this? I'd like to see we're aligned with how the monitoring team is gathering this metric.

pstefek-mongo: I will check! When you say this do you mean they already are recording this metric in mongonet? If so what is it called? we can probably just hijack that

pstefek-mongo: update: it looks like the monitoring team is not interested in this metric so we are not currently gathering it

proxy_session.go
pstefek-mongo: Cool I think it should be a separate pr but I'll file a ticket to remove this interface

PR #95, opened 2021-09-02
CLOUDP-98421 (followup) Pass operation type to response interceptor and add total time taken metric

pstefek-mongo:
Jira: https://jira.mongodb.org/browse/CLOUDP-98421 also covers https://jira.mongodb.org/browse/CLOUDP-98878
EVG:  https://evergreen.mongodb.com/version/613115ba3627e0298b8f5bfa?redirect_spruce_users=true


TomerYakir:
hmm.. so we're accumulating the proxy processing time but not the db processing time? would that be a problem when comparing the metrics?

pstefek-mongo:
Great point @TomerYakir db time tracking is now cumulative

TomerYakir (approved):

PR #98, opened 2021-09-28
Close connection after RunCommandUsingRawBSON

ahmedssulaiman:
Close connection after RunCommandUsingRawBSON to prevent connection leaks

Ran tests with and without change - verified test only passes when we close the connection


ahmedssulaiman:
While I can monitor connections using lsof, not sure what's the best way to reliably write an integration test for this. 
I am also thinking of running the ping twice in `TestCommonRunCommandUsingRawBSON` to check that the results are consistent between calls.

TomerYakir:
When I manually tested it, I was checking serverStatus().connections to monitor open connections.

ahmedssulaiman:
Any reason why https://github.com/mongodb/mongonet/blob/master/inttests/run_integration_tests_mongod_mode.sh does not include `go test -test.v -run TestCommon` ?

I see it added in #81 but not in #82

inttests/mongoapi_test.go
```diff
@@ -34,5 +36,31 @@ func TestCommonRunCommandUsingRawBSON(t *testing.T) {
...
+
+	version, err := serverStatus.LookupErr("version")
+	if err != nil {
+		t.Fatalf("Error fetching version from serverStatus")
+	}
+
+	if util.CompareVersions(version.StringValue(), "5.0") < 0 {
+		t.Log("serverStatus.connections, only available in 5.0 and above")
+		return 0
+	}
```
TomerYakir, line 60: I'm pretty sure serverStatus had connections pre 5.0

TomerYakir:
> Any reason why https://github.com/mongodb/mongonet/blob/master/inttests/run_integration_tests_mongod_mode.sh does not include `go test -test.v -run TestCommon` ?
> 
> I see it added in #81 but not in #82

Might've been an oversight. It should have TestCommon in both

inttests/mongoapi_test.go
ahmedssulaiman: ah nvm, will remove this check then. misread the docs: https://docs.mongodb.com/manual/reference/command/serverStatus/#connections
Seems like only some fields are new in 5.0

ahmedssulaiman:
Seems like it might have been intentional, `TestCommonProxyProtocolInt` and `TestCommonPinnedServer ` both fail in mongod mode

ahmedssulaiman:
Will rename those to `TestProxyMongosMode` since they are only meant for mongos mode

inttests/mongoapi_test.go
```diff
@@ -34,5 +36,21 @@ func TestCommonRunCommandUsingRawBSON(t *testing.T) {
 	if v != 1 {
 		t.Fatalf("expected to get ok:1 but got %v", v)
 	}
+	newNumberOfConnections := GetNumberOfConnections(t, client, goctx)
```
TomerYakir, line 39: [q/ignorable] do we want to run `RunCommandUsingRawBSON` multiple times?
I'm fine not doing this as long as you can confirm that the test fails without the `defer conn.Close()` change

inttests/mongoapi_test.go
```diff
@@ -34,5 +36,21 @@ func TestCommonRunCommandUsingRawBSON(t *testing.T) {
...
+	if numberOfConnections != newNumberOfConnections {
+		t.Fatalf("connection leak, expected numberOfConnections to be %v, got %v", numberOfConnections, newNumberOfConnections)
+	}
+}
+
+func GetNumberOfConnections(t *testing.T, client *mongo.Client, ctx context.Context) int32 {
+	serverStatus, err := client.Database("admin").RunCommand(
```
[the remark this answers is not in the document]
ahmedssulaiman: I'll add a check.

inttests/mongoapi_test.go
ahmedssulaiman: I did verify the test fails without this change (number of connections went up from 3->4) but there is no harm in running this a few times (e.g. 5)

inttests/mongoapi_test.go
```diff
@@ -19,26 +19,28 @@ func TestCommonRunCommandUsingRawBSON(t *testing.T) {
...
-		t.Fatalf("expected to that 'ok' is int but got %T", raw)
-	}
-	if v != 1 {
-		t.Fatalf("expected to get ok:1 but got %v", v)
-	}
-	newNumberOfConnections := GetNumberOfConnections(t, client, goctx)
-	if numberOfConnections != newNumberOfConnections {
-		t.Fatalf("connection leak, expected numberOfConnections to be %v, got %v", numberOfConnections, newNumberOfConnections)
+	for i := 0; i < 5; i++ {
+		numberOfConnections := GetNumberOfConnections(t, client, goctx)
```
TomerYakir, line 23: [q] should we get the original numberOfConnections before the loop? (and newNumberOfConnections after the loop?)

TomerYakir (approved):
LGTM on green!

PR #99, opened 2021-12-03
Add support for parsing private endpoint TLV information

d-kupiec:
Add support for parsing private endpoint information from proxy protocol v2 TLVs, as defined by the spec [1].
Private endpoint id TLV encoding for AWS, Azure, and GCP can be found at [2][3][4]
Pass these ids to the session and expose via a new GetPrivateEndpointId().

[1] http://www.haproxy.org/download/1.8/doc/proxy-protocol.txt
[2] https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-target-groups.html#proxy-protocol
[3] https://docs.microsoft.com/en-us/azure/private-link/private-link-service-overview#getting-connection-information-using-tcp-proxy-v2
[4] https://cloud.google.com/vpc/docs/configure-private-service-connect-producer#proxy-protocol


connwrapper.go
```diff
@@ -279,6 +310,51 @@ func (c *Conn) initv2() error {
...
+		}
+		tlvPayload := payload[nextByte+3 : nextByte+3+tlvLength]
+		nextByte += 3 + tlvLength
+		switch tlvType {
+		case PP2_TYPE_AWS:
+			// currently only 0x01 subtype is supported
+			if tlvPayload[0] == 0x01 {
+				c.privateEndpointInfo = PrivateEndpointInfo{
+					base64.StdEncoding.EncodeToString([]byte{PP2_TYPE_AWS}),
+					base64.StdEncoding.EncodeToString(tlvPayload[1:]),
```
TomerYakir, line 333: [q] are we certain `tlvPayload[1:]` can't fail with out of bound index?

TomerYakir:
This LGTM, but I wonder if it's worth adding a fuzz test to the function that parses the proxy protocol to ensure we don't panic on any malformed messages.

connwrapper.go
d-kupiec: Added additional checks to prevent indexing into the subtype in the event that the TLV payload length is unexpectedly 0.

d-kupiec:
> This LGTM, but I wonder if it's worth adding a fuzz test to the function that parses the proxy protocol to ensure we don't panic on any malformed messages.

Looks like fuzzing is coming natively to golang in 1.18 (currently in beta). In the meantime, is there a recommended package for setting up fuzz tests in golang?

TomerYakir:
We were using previously using https://github.com/dvyukov/go-fuzz for our other project

d-kupiec:
> We were using previously using https://github.com/dvyukov/go-fuzz for our other project

Will leave adding a fuzzer as a follow up and going ahead with this submission now.

PR #104, opened 2022-03-07
Warm connection pool in perf tests

TomerYakir:


inttests/testrunner.go
```diff
@@ -20,6 +20,75 @@ import (
 	"go.mongodb.org/mongo-driver/mongo"
 )
 
+func doIteration(logger *slogger.Logger, hostname string, proxyPort int, mode util.MongoConnectionMode,
```
TomerYakir, line 23: [note] this is mostly extracting the iteration loop into a function

inttests/testrunner.go
```diff
@@ -20,6 +20,75 @@ import (
...
+	var client *mongo.Client
+	ctx := context.Background()
+	client, err = proxyClientFactory(hostname, proxyPort, mode, false, "perftest", context.Background())
+	if err != nil {
+		logger.Logf(slogger.ERROR, "failed to init connection for cleanup. err=%v", err)
+		return []int64{}, err
+	}
+	defer client.Disconnect(ctx)
+	var wg sync.WaitGroup
+	for i := 0; i < workers*2; i++ {
```
TomerYakir, line 40: [note] this is an actual change

inttests/testrunner.go
```diff
@@ -157,7 +188,7 @@ ITERATIONS:
 	percentiles[80] = sortedResults[int(float32(len(sortedResults))*0.8)]
 	percentiles[90] = sortedResults[int(float32(len(sortedResults))*0.9)]
 	percentiles[95] = sortedResults[int(float32(len(sortedResults))*0.95)]
-	percentiles[99] = sortedResults[int(float32(len(sortedResults))*0.95)]
+	percentiles[99] = sortedResults[int(float32(len(sortedResults))*0.99)]
```
TomerYakir, line 191: [note] drive-by fix

inttests/testrunner.go
```diff
@@ -262,8 +293,8 @@ func analyzeResults(err error, workers int, failedCount int32, avgLatencyMs, tar
 	if avgLatencyMs > targetAvgLatencyMs {
 		return fmt.Errorf("ERROR average latency %v > %v", avgLatencyMs, targetAvgLatencyMs)
 	}
-	if maxLatencyMs > targetMaxLatencyMs {
-		return fmt.Errorf("ERROR max latency %v > %v", maxLatencyMs, targetMaxLatencyMs)
+	if int64(percentiles[95]) > targetMaxLatencyMs {
```
TomerYakir, line 296: [note] using 95-ile instead of max to reduce test sensitivity to outliers

inttests/testrunner.go
```diff
@@ -84,52 +153,14 @@ func DoConcurrencyTestRun(logger *slogger.Logger,
...
-			logger.Logf(slogger.INFO, "*** iteration %v has failures. breaking", j)
-			break ITERATIONS
+		var results2 []int64
+		results2, err = doIteration(logger, hostname, proxyPort, mode, workers, j, &failedCount, proxyClientFactory, testFunc)
+		if err != nil {
+			logger.Logf(slogger.ERROR, "%v", err)
+			return
 		}
-		time.Sleep(500 * time.Millisecond)
+		results = append(results, results2...)
```
d-kupiec, line 163: Thoughts on keeping the sleep here rather than having doIteration responsible for sleeping?

inttests/testrunner.go
```diff
@@ -20,6 +20,75 @@ import (
...
+	workers int,
+	iter int,
+	failedCount *int32,
+	proxyClientFactory func(host string, port int, mode util.MongoConnectionMode, secondaryReads bool, appName string, ctx context.Context) (*mongo.Client, error),
+	testFunc func(logger *slogger.Logger, client *mongo.Client, workerNum, iteration int, ctx context.Context) (elapsed time.Duration, success bool, err error),
+) (results []int64, err error) {
+	logger.Logf(slogger.INFO, "*** starting iteration %v", iter)
+	var client *mongo.Client
+	ctx := context.Background()
+	client, err = proxyClientFactory(hostname, proxyPort, mode, false, "perftest", context.Background())
```
d-kupiec, line 33: Just to confirm this change in behavior, it looks like previously for each iteration, each worker would have it's own client connection, and now all the workers share the same client connection (without ClientTimeoutSecForTests)

inttests/testrunner.go
TomerYakir: 👍

inttests/testrunner.go
TomerYakir: correct - the reason is that we'd like to have the connection pool warmed up for each iteration

d-kupiec (approved):
lgtm!

PR #106, opened 2022-05-12
Fix BSONWalkAll unexpected REMOVE_FIELD behavior

NameThatFollows:
Running `BSONWalkAll` with a visitor that returns a `REMOVE_FIELD` resulted in some unexpected behavior:

- The last element in the resulting BSON would be duplicated (new slice was not assigned to the element, so existing slice would still show the same number of elements)
- Removing a field would cause the walk to short-circuit, preventing candidates in further iterations from being seen. 

This also adds a test to cover these two cases.


TomerYakir:
LGTM - adding @d-kupiec for another pair of eyes

TomerYakir (approved):

bsonutil.go
```diff
@@ -209,10 +210,13 @@ func BSONWalkAll(doc bson.D, fieldName string, visitor BSONWalkVisitor) (bson.D,
 				if err == DELETE_ME {
 					return nil, nil
```
d-kupiec, line 211: By similar logic, does DELETE_ME delete the entire given bson or just the element in the current doc that generated DELETE_ME on Visit()?

bsonutil.go
```diff
@@ -209,10 +210,13 @@ func BSONWalkAll(doc bson.D, fieldName string, visitor BSONWalkVisitor) (bson.D,
...
 				}
+				// if one field needs to be removed, store the index, and continue processing remaining
+				// fields/documents
 				if err == REMOVE_FIELD {
-					return append(current[:i], current[i+1:]...), nil
+					docRemovedIndex = i
```
d-kupiec, line 216: If the field is marked for removal, perhaps we should still short circuit the processing below and continue to the next element

bsonutil.go
```diff
@@ -254,6 +258,9 @@ func BSONWalkAll(doc bson.D, fieldName string, visitor BSONWalkVisitor) (bson.D,
 		}
 
 	}
+	if docRemovedIndex > -1 {
```
d-kupiec, line 261: Do we expect to only encounter one element resulting in REMOVE_FIELD per call, or might we see more than one now that we arent preemptively returning on the first one?

d-kupiec (commented):
a few questions as I learn more about this particular area of code and the expectations for it

bsonutil.go
NameThatFollows: I believe the former, DELETE_ME deletes the entire bson that contains the field.

d-kupiec: Got it! Would you mind updating the comment above to rephrase from 'that element' to 'the BSON doc' for clarity?

bsonutil.go
NameThatFollows: I assumed that a document (excluding sub-documents) will only ever have one instance of a given field/key. Is it a valid use-case to have a bson with multiple fields with the same field name? If so (or just to be safe), I can probably modify this to track multiple removed fields.

bsonutil.go
NameThatFollows: Good point! Let me try out that change

bsonutil.go
d-kupiec: It looks like for BSON, it is possible a document may have more than one field with the same name (though unlikely), https://www.mongodb.com/docs/manual/core/document/#field-names

Not sure if Tomer has some more insight regarding handling of this scenario, though I imagine for a base level util we probably want to support it if our docs specify it's allowed

bsonutil.go
```diff
@@ -198,29 +198,29 @@ type BSONWalkVisitor interface {
 }
 
 // BSONWalkAll - recursively walks the BSON doc and applies the visitor when encountering the fieldName
-// If delete_me is encountered, it'll return an empty document for that element
+// If delete_me is encountered, it'll return an empty document for that BSON doc
+// If remove_field is encountered, it'll return the BSON doc with the field ommitted
```
d-kupiec, line 202: nit: s/ommitted/omitted

bsonutil.go
```diff
@@ -258,10 +258,11 @@ func BSONWalkAll(doc bson.D, fieldName string, visitor BSONWalkVisitor) (bson.D,
 		}
 
 	}
-	if docRemovedIndex > -1 {
-		return append(current[:docRemovedIndex], current[docRemovedIndex+1:]...), nil
+	for i := len(docsToRemove) - 1; i >= 0; i-- {
+		current = append(current[:i], current[i+1:]...)
```
d-kupiec, line 262: I think we want docsToRemove[i]

bsonutil_test.go
```diff
@@ -514,6 +526,57 @@ func TestBSONWalkAll7(test *testing.T) {
...
+	}
+	walker := &removeWalker{}
+	doc, err := BSONWalkAll(doc, "a", walker)
+	if err != nil {
+		test.Errorf("why did we get an error %s", err)
+	}
+	if len(walker.seen) != 2 {
+		test.Errorf("wrong # saw %d", len(walker.seen))
+	}
+	if len(doc) != 1 {
```
d-kupiec, line 543: Perhaps assert the remaining element is b to ensure the right elements are being removed.

bsonutil.go
NameThatFollows: 🤦 thanks for catching that!

d-kupiec (approved):
lgtm, thanks!

PR #107, opened 2022-05-17
CLOUDP-122235 - Add gRPC server support

denniskuczynski:
This PR does some refactoring to support a new underlying server type: GRPCServer alongside the existing server renamed TCPServer.
Currently the server does not actually process data or create sessions.

TCPSession and GRPCSession will be extracted into a new Session interface for processing by ProxySession.

There should not be any API breaking changes outside of the changes to the Proxy and ProxyConfig structures to add the new fields. 

This PR also adds the grpc bytes codec as it was necessary for the minimal test case.


config.go
```diff
@@ -50,18 +53,24 @@ func NewProxyConfig(bindHost string, bindPort int, mongoUri, mongoHost string, m
...
+			0,   // TCPKeepAlivePeriod
+			nil, // CipherSuites
 		},
+		GRPCServerConfig{
+			false,
+			"",
+			0,
+		},
+		slogger.OFF, // LogLevel
+		nil,         // Appenders
```
denniskuczynski, line 73: log configuration settings now at the root.  LoggerFactory's are passed to the individual servers

proxy_session.go
```diff
@@ -22,8 +22,7 @@ import (
 )
 
 type ProxySession struct {
-	*Session
-
+	session          Session
```
denniskuczynski, line 25: oh, I just realized that changing this means `ps.SSLServerName` won't work.  Thinking if I should push the unqiue Sessions down a layer or make SSLServerName a function...

grpc.go
```diff
@@ -0,0 +1,23 @@
...
+type BytesCodec struct{}
+
+// Dereference []byte slice pointer
+func (c BytesCodec) Marshal(v interface{}) ([]byte, error) {
+	rawMessage := v.(*[]byte)
+	return *rawMessage, nil
+}
+
+// Expect v to be empty []byte slice pointer
+func (c BytesCodec) Unmarshal(data []byte, v interface{}) error {
```
TomerYakir, line 15: [q] could we add fuzz tests for this one?

proxy_session.go
```diff
@@ -126,77 +126,6 @@ func (ps *ProxySession) DoLoopTemp() {
 	}
 }
 
-func (ps *ProxySession) respondWithError(clientMessage Message, err error) error {
```
TomerYakir, line 129: [q] is that unused code?

denniskuczynski: yes, it does not seemed to be called anywhere

TomerYakir:
LGTM - ping me when it's ready for review!

grpc.go
denniskuczynski: I added test cases in grpc_test.go.  I played around with adding a Go 1.18 fuzz test, but I don't think this method is interesting enough for fuzzing considering the time that adds.

denniskuczynski:
This should be ready to review. Note that this will just be going into the grpc feature branch.
cc @NataliaSulzhenko

go.sum
```diff
@@ -1,7 +1,26 @@
+cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
TomerYakir, line 1: [q] have we run `go mod tidy`?

denniskuczynski: updated https://github.com/mongodb/mongonet/pull/107/commits/caa514a3f4d7d9da3ff343dd02d2622773756113

TomerYakir (approved):
LGTM - since we're merging to a feature branch it doesn't run the automated tests. Have we confirmed those are passing?

go.mod
```diff
@@ -6,10 +6,12 @@ require (
 	github.com/go-test/deep v1.0.8
 	github.com/mongodb/slogger v0.0.0-20200908173206-e26499443662
 	go.mongodb.org/mongo-driver v1.7.2-0.20210804202952-82d3467025bc // pinned version
+	google.golang.org/grpc v1.46.0
 )
 
 require (
 	github.com/go-stack/stack v1.8.1 // indirect
+	github.com/golang/protobuf v1.5.2 // indirect
```
NataliaSulzhenko, line 14: Dennis, this might be a silly question, but do we have this dependency because of grpc lib? does it require it internally by default? same for `google.golang.org/genproto` and `google.golang.org/protobuf`

proxy_session.go
```diff
@@ -85,7 +85,11 @@ func (ps *ProxySession) GetLogger() *slogger.Logger {
 }
 
 func (ps *ProxySession) ServerPort() int {
-	return ps.proxy.Config.BindPort
+	if ps.IsTCPSession() {
+		return ps.proxy.Config.TCPServerConfig.BindPort
```
NataliaSulzhenko, line 89: @denniskuczynski [q] I believe for gRPC we also should return bind port, right?

denniskuczynski: Yes, I haven't added any gRPC session related code, so we just return 0 for now if not TCP.  But I will change this in a follow-on PR.

denniskuczynski:
@TomerYakir yes, I confirmed integration (mongod/mongos mode) and unit tests pass locally.

PR #108, opened 2022-05-25
CLOUDP-122238 - Handle incoming requests on gRPC channel using ProxySessions

denniskuczynski:
This adds basic gRPC session handling by adapting the existing SendMessage/ReadMessage functions to leverage a gRPC stream instead of a TCP socket for a session.

server_test.go adds a basic sanity test with its mocked ServerTestFactory
```
TestGRPCServer (0.00s)
    server_test.go:421: Sending gRPC msg: [{insert foo} {$db test} {documents [[{_id 0}]]}]
    server_test.go:466: Received gRPC response: [{ok 1}]
    server_test.go:421: Sending gRPC msg: [{find foo} {$db test} {filter []}]
    server_test.go:466: Received gRPC response: [{cursor [{firstBatch [[{foo 17}]]} {id 0} {ns test.foo}]} {ok 1}]
```


server_test.go
```diff
@@ -230,10 +231,11 @@ func (mss *MyServerSession) Close() {
 }
 
 type MyServerTestFactory struct {
+	mydata map[string][]bson.D
```
denniskuczynski, line 234: lifted the data map up the the factory level as gRPC sessions do not live as long as TCP sessions

server.go
```diff
@@ -360,16 +352,14 @@ func (s *GRPCServer) Run() error {
 						md,
 					)
 
-					var in []byte
-					if err := stream.RecvMsg(&in); err != nil {
-						return fmt.Errorf("Unable to read message from client")
+					var serverName string
+					serverNameMD := md.Get("ServerName")
+					if len(serverNameMD) == 1 {
```
TomerYakir, line 357: [ignorable/q] do we need to handle other cases? len != 1

server.go
```diff
@@ -360,16 +352,14 @@ func (s *GRPCServer) Run() error {
...
+						serverName = serverNameMD[0]
 					}
 
-					var out []byte
-					if err := stream.SendMsg(&out); err != nil {
-						return fmt.Errorf("Unable to send response to client")
-					}
-					return nil
+					c := NewGRPCSession(s, serverName, peer.Addr)
```
TomerYakir, line 361: [q] is peer.Addr the Envoy IP or the remote client IP?

server_test.go
```diff
@@ -373,6 +376,34 @@ func TestGRPCServer(t *testing.T) {
...
+	err = sendWithGRPC(t, conn, find)
+	if err != nil {
+		t.Error(err)
+	}
+
+	// Shutdown
+	server.GracefulStop()
+}
+
+func sendWithGRPC(t *testing.T, conn *grpc.ClientConn, bson bson.D) error {
```
TomerYakir, line 406: [nit/ignorable] could we rename so it's clear this function does a roundtrip?

TomerYakir (approved):
LGTM 💯

session.go
```diff
@@ -121,6 +159,18 @@ func (s *Session) Run(conn net.Conn) {
...
 
+// Run executes a GRPCSession with a grpc.ServerStream
+func (s *Session) RunWithStream(stream grpc.ServerStream) error {
+	s.stream = stream
+	worker, err := s.grpcServer.workerFactory.CreateWorker(s)
+	if err != nil {
+		s.logger.Logf(slogger.WARN, "error creating worker %v", err)
+		return err
+	}
+	worker.DoLoopTemp()
```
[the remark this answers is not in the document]
denniskuczynski: yes, my previous handling code, looked like:
```go
for {
  var out []byte
  err := stream.RecvMsg(&out)
  if err == io.EOF {
    break
  }
  ...
```

server.go
denniskuczynski: I think this will probably be the envoy address depending on how that's setup. So this might have to change eventually.
https://pkg.go.dev/google.golang.org/grpc/peer

server.go
denniskuczynski: yeah, we might want to make this an error in the future.  I wasn't sure if the error should be in the mongonet layer.

server_test.go
denniskuczynski: renamed this method: https://github.com/mongodb/mongonet/pull/108/commits/1f8443229b2e8a45dc076abe8263130233d32b49

denniskuczynski:
Added a sanity integration test as well: https://github.com/mongodb/mongonet/pull/108/commits/8b98f076ef28bed05f79b5b673748c5d83037d5c

inttests/proxy_grpc_sanity_test.go
```diff
@@ -0,0 +1,149 @@
+package inttests
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/mongodb/mongonet"
+	. "github.com/mongodb/mongonet"
```
TomerYakir, line 10: [q] duplicate dependency?

inttests/proxy_grpc_sanity_test.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/mongodb/slogger/v2/slogger"
+	"go.mongodb.org/mongo-driver/bson"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+func privateSanityTestGRPC(t *testing.T) {
+	mongoPort, proxyPort, hostname := util.GetTestHostAndPorts()
+	t.Logf("using proxy port %v", proxyPort)
+	pc := getProxyConfig(hostname, mongoPort, proxyPort, DefaultMaxPoolSize, DefaultMaxPoolIdleTimeSec, util.Direct, false, nil)
```
TomerYakir, line 21: [q] is "Cluster-mode" applicable for this test?

inttests/proxy_grpc_sanity_test.go
```diff
@@ -0,0 +1,149 @@
...
+		{"find", "foo"},
+		{"$db", "test"},
+		{"filter", bson.D{}},
+	}
+	err = sendAndRecvMsg(t, conn, find)
+	if err != nil {
+		t.Error(err)
+	}
+
+	defer conn.Close()
```
TomerYakir, line 69: [q] would we want to move this defer statement right after the `grpc.Dial` error handling statement?

inttests/proxy_grpc_sanity_test.go
denniskuczynski: not currently.  The Mongod mode integration tests connect directly to a single mongod process.

inttests/proxy_grpc_sanity_test.go
denniskuczynski: oops, bad copy paste

denniskuczynski:
@TomerYakir Updated 826d0c3a5be24ec4a08dd7ab0291ed5861236dab

TomerYakir (approved):

PR #112, opened 2022-06-22
CLOUDP-122238 - Add support for pinning gRPC sessions using transactions

denniskuczynski:
This PR adds support for pinning sessions by transaction when using gRPC sessions.

When we receive the gRPC request we need to read the initial message outside of the main loop to determine if we might have a pinned session.  We then pass this read method to the ServerWorker's DoLoopTemp method for further processing.

Pinned transactions are keyed by the lsid (session identifier `bson.D` UUID) and the txnNumber (`int64`).  I also added cursorId to this lookup structure as we should be able to reuse as any cursors within a transaction would be using the transaction's pinning.

The inttest shows the basic API of:
* mongonet handling pinned session lookups based on the OpMsg cmd fields
* proxy implementation code responsible for triggering pinned sessions via `InterceptClientToMongo` return value. 
* proxy implementation code responsible for unpinning sessions via session `UnpinByTransaction` method.


server.go
```diff
@@ -309,22 +312,126 @@ func (s *TCPServer) contextualWorkerFactory() (ServerWorkerWithContextFactory, b
...
+	defer s.pinnedSessionMapLock.Unlock()
+
+	s.pinnedSessionMap[&pinnedSessionKey{
+		SSLServerName: SSLServerName,
+		lsid:          txn.lsid,
+		txnNumber:     txn.txnNumber,
+	}] = worker
+}
+
+func (s *GRPCServer) PinSessionByCursorId(worker ServerWorker, SSLServerName string, cursorId int64) {
```
NataliaSulzhenko, line 363: @denniskuczynski thank you :)

server.go
```diff
@@ -309,22 +312,126 @@ func (s *TCPServer) contextualWorkerFactory() (ServerWorkerWithContextFactory, b
...
+}
+
+func (s *GRPCServer) LookupSessionByTransaction(SSLServerName string, lsid bson.D, txnNumber int64) (ServerWorker, bool) {
+	s.pinnedSessionMapLock.RLock()
+	defer s.pinnedSessionMapLock.RUnlock()
+
+	for k, v := range s.pinnedSessionMap {
+		if k.SSLServerName == SSLServerName &&
+			k.txnNumber == txnNumber &&
+			reflect.DeepEqual(k.lsid, lsid) {
```
NataliaSulzhenko, line 380: @denniskuczynski I need to read more about reflection in go, but I'm a bit concerned about performance in here. 
wondering if it's possible to write a custom func (something like comparator :) )

PR #114, opened 2022-07-13
CLOUDP-125897: Store sticky ProxySessions for subsequent reads from cursors, using a composite key: proxyCursorId and tenantId.

NataliaSulzhenko:


proxy_session.go
```diff
@@ -47,7 +47,7 @@ type MetricsHookFactory interface {
 }
 
 type ResponseInterceptor interface {
-	InterceptMongoToClient(m Message, serverAddress address.Address, isRemote bool, retryAttemptsExhausted bool, operationType string) (Message, error)
+	InterceptMongoToClient(m Message, serverAddress address.Address, isRemote bool, retryAttemptsExhausted bool, operationType string) (int64, Message, error)
```
denniskuczynski, line 50: I think in the spec doc we said:
`(Message, proxyCursorId int64, error)` as the return type.  I think that might make more sense to have Message first.

denniskuczynski (commented):
One nit from me but otherwise LGTM 👍

proxy_session.go
NataliaSulzhenko: good catch, updating

denniskuczynski (approved):

PR #116, opened 2022-07-25
CLOUDP-108151 - Add support for gRPC server version of TCP wire protocol

denniskuczynski:
This PR adds support for a gRPC server version of the MongoDB wire protocol to mongonet.

It achieves this using the https://github.com/grpc/grpc-go library to start an embedded gRPC server using a custom bytes codec.

gRPC server configuration is now separated from TCP server configuration as well.

This has the following breaking API changes:
* Configuration structs have had fields renamed and moved to distinguish between gRPC and TCP configuration.
* Interceptor APIs have been modified to return more information about pinned transactions and pinned cursor IDs for sessions.

The version of mongonet is still at v0.0.0, so these breaking changes should be fine for any existing clients now per the rules of semantic versioning.

The inttest changes and unit test changes in server_test.go demonstrate the new APIs.

See also https://github.com/mongodb/mongonet/pull/115 for a periodic session reaper goroutine. This may be useful for cleaning up memory, but is not strictly necessary for now.


proxy_session.go
```diff
@@ -126,77 +145,6 @@ func (ps *ProxySession) DoLoopTemp() {
 	}
 }
 
-func (ps *ProxySession) respondWithError(clientMessage Message, err error) error {
```
denniskuczynski, line 129: unused

inttests/proxy_grpc_sanity_test.go
```diff
@@ -0,0 +1,315 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	proxy.InitializeServer()
+	if ok, _, _ := proxy.OnSSLConfig(nil); !ok {
+		panic("failed to call OnSSLConfig")
+	}
+
+	go proxy.Run()
```
TomerYakir, line 39: [q] do we want to add a short sleep here, just in case?

inttests/proxy_grpc_sanity_test.go
```diff
@@ -0,0 +1,315 @@
...
+		}
+	}
+	if trailer != nil {
+		t.Logf("Received trailer: %v", trailer)
+	}
+	return nil
+}
+
+func TestProxyMongosModeGRPCSanity(t *testing.T) {
+	privateSanityTestGRPC(t)
```
TomerYakir, line 160: [q] any reason why this is implemented as a "private" function? I think that starting with go1.18, `go vet` starts complaining about passing `t` to non-test functions

session.go
```diff
@@ -59,13 +99,56 @@ func (s *Session) SetPrivateEndpointId(id string) {
 	s.privateEndpointInfo.privateEndpointId = id
 }
 
+func (s *Session) IsSticky() bool {
+	return s.sticky
+}
+
+func (s *Session) SetSticky(sticky bool) {
```
TomerYakir, line 106: [q] do we ever set this to `true`?

server.go
```diff
@@ -314,32 +307,309 @@ func NewServer(config ServerConfig, factory ServerWorkerFactory) Server {
...
+}
+
+func (s *GRPCServer) LookupSessionByTransaction(SSLServerName string, lsid bson.D, txnNumber int64) (ServerWorker, bool) {
+	s.pinnedSessionMapLock.RLock()
+	defer s.pinnedSessionMapLock.RUnlock()
+
+	for k, v := range s.pinnedSessionMap {
+		if k.SSLServerName == SSLServerName &&
+			k.txnNumber == txnNumber &&
+			reflect.DeepEqual(k.lsid, lsid) {
```
TomerYakir, line 380: [q/ignorable] are we concerned about any performance penalty here for using reflection? have we considered somehow hashing the `lsid` value on the write path to optimize the reads?
Marking as ignorable though since we don't expect a large amount of transactions for a given SNI/txnNumber combination.

session.go
```diff
@@ -91,18 +174,49 @@ func (s *Session) Run(conn net.Conn) {
...
+	worker.DoLoopTemp(m)
+	if !s.IsSticky() {
+		s.logger.Logf(slogger.DEBUG, "Closing current worker session")
+		worker.Close()
+	}
+	return nil
+}
+
+// UnpinByTransaction should be called by user code to unpin a pinned session for transactions
+func (s *Session) UnpinByTransaction(lsid bson.D, txnNumber int64) bool {
```
TomerYakir, line 202: [q] do we ever use the return value of this function anywhere in the code?

session.go
```diff
@@ -91,18 +174,49 @@ func (s *Session) Run(conn net.Conn) {
...
+	}
+	found := s.grpcServer.UnpinSessionByTransaction(s.SSLServerName, lsid, txnNumber)
+	if found {
+		s.Logf(slogger.DEBUG, "Unpinned session by transaction %v/%v", lsid, txnNumber)
+		s.SetSticky(false)
+	}
+	return found
+}
+
+// UnpinByTransaction should be called by user code to unpin a pinned session for cursors
```
TomerYakir, line 214: [sp] should be UnpinByCursorId

session.go
```diff
@@ -91,18 +174,49 @@ func (s *Session) Run(conn net.Conn) {
...
+func (s *Session) UnpinByCursorId(cursorId int64) bool {
+	if !s.IsGRPCSession() {
+		return false
+	}
+	return s.grpcServer.UnpinSessionByCursorId(s.SSLServerName, cursorId)
```
TomerYakir, line 219: [q] should logging around this be consistent with `UnpinByTransaction`?

server.go
```diff
@@ -314,32 +307,309 @@ func NewServer(config ServerConfig, factory ServerWorkerFactory) Server {
...
+						md,
+					)
+
+					// Attempt to specify TLS ServerName field via default metadata field name
+					var serverName string
+					serverNameMD := md.Get("ServerName")
+					if len(serverNameMD) == 1 {
+						serverName = serverNameMD[0]
+					} else {
+						s.logger.Logf(slogger.WARN, "Expected ServerName incoming metadata was not present")
```
TomerYakir, line 468: [q] what would it mean if we ever get such connections with no ServerName metadata? is there a point in carrying on handling such connection?

server.go
```diff
@@ -314,32 +307,309 @@ func NewServer(config ServerConfig, factory ServerWorkerFactory) Server {
...
+								slogger.WARN,
+								"Error resolving RemoteAddr metadata: %v %v",
+								remoteAddrMD[0],
+								err,
+							)
+						} else {
+							remoteAddr = resolvedAddr
+						}
+					} else {
+						s.logger.Logf(slogger.WARN, "Expected RemoteAddr incoming metadata was not present")
```
TomerYakir, line 488: [q] is it possible to ever see a `len` > 1?

server.go
```diff
@@ -314,32 +307,309 @@ func NewServer(config ServerConfig, factory ServerWorkerFactory) Server {
...
+			},
+		},
+	}, nil)
+
+	lis, err := net.Listen("tcp", fmt.Sprintf("%v:%d", s.config.BindHost, s.config.BindPort))
+	if err != nil {
+		return err
+	}
+
+	s.logger.Logf(slogger.DEBUG, "Starting local gRPC server %v\n", lis.Addr())
```
TomerYakir, line 526: [q] the TCP server uses `slogger.WARN` for the equivalent message - do we want to do the same here?
(actually, I'm not sure why we didn't log at INFO-level in the first place)

server.go
```diff
@@ -314,32 +307,309 @@ func NewServer(config ServerConfig, factory ServerWorkerFactory) Server {
...
+	case *MessageMessage:
+		bodyDoc, err := message.BodyDoc()
+		if err != nil {
+			return nil, nil, err
+		}
+		simpleBSON, err := SimpleBSONConvert(bodyDoc)
+		if err != nil {
+			return nil, nil, err
+		}
+		cmd, err := simpleBSON.ToBSOND()
```
TomerYakir, line 557: [q] (this probably be handled in a follow-up PR) given that a) our application would also convert the message to `bson.D` and b) this is on the hot path and c) it's a relatively expensive operation - should we also return the `bson.D` to downstream consumers?

server_test.go
```diff
@@ -349,6 +355,245 @@ func TestServerWorkerWithContext(t *testing.T) {
 	}
 }
 
+func TestGRPCServer(t *testing.T) {
+	port := 9920 // TODO: pick randomly or check?
```
TomerYakir, line 359: [misc] I think that this `TODO` can be removed at this point (also from the other occurrence)

server_test.go
```diff
@@ -349,6 +355,245 @@ func TestServerWorkerWithContext(t *testing.T) {
...
+			"127.0.0.1",
+			port,
+		},
+		&MyServerTestFactory{mydata: map[string][]bson.D{}},
+		func(prefix string) *slogger.Logger {
+			return mongonet.NewLogger(prefix, slogger.DEBUG, []slogger.Appender{slogger.StdOutAppender()})
+		},
+	)
+
+	go server.Run()
```
TomerYakir, line 372: (same comment on adding a short sleep)

server_test.go
```diff
@@ -349,6 +355,245 @@ func TestServerWorkerWithContext(t *testing.T) {
...
+		{"autocommit", false},
+		{"txnNumber", int64(1)},
+		{"lsid", bson.D{
+			{"id", id},
+		}},
+		{"documents", bson.A{
+			bson.D{{"_id", int64(0)}},
+		}},
+	}
+	err = sendAndRecvMsg(t, conn, insert, 251)
```
TomerYakir, line 456: [q] could we use a constant for `251`?

server.go
```diff
@@ -314,32 +307,309 @@ func NewServer(config ServerConfig, factory ServerWorkerFactory) Server {
...
+				txnNumberIdx := BSONIndexOf(cmd, "txnNumber")
+				if txnNumberIdx >= 0 {
+					txnNumber, ok := cmd[txnNumberIdx].Value.(int64)
+					if !ok {
+						return nil, nil, fmt.Errorf("txnNumber was not of type int64")
+					}
+					s.logger.Logf(slogger.DEBUG, "Looking up session by transaction %v %v/%v", SSLServerName, lsid, txnNumber)
+					worker, found := s.LookupSessionByTransaction(SSLServerName, lsid, txnNumber)
+					if !found {
+						return nil, m, NewMongoError(fmt.Errorf("Given transaction number %v does not match any in-progress transactions. The active transaction number is -1", txnNumber), 251, "NoSuchTransaction")
```
TomerYakir, line 583: [nit] could we use a constant for 251? (and 43 below)

TomerYakir:
Looks like there's a failed unit test

inttests/proxy_grpc_sanity_test.go
denniskuczynski: I had changed the grpc.Dial function to use grpc.WithBlock() which should account for the server start up time.  I probably should add an explicit timeout there though.

inttests/proxy_grpc_sanity_test.go
denniskuczynski: no I was copying the style in `proxy_sanity_test.go` so we can change that.

session.go
denniskuczynski: yes, I'll change

session.go
denniskuczynski: We do not.  I can change this API

server.go
denniskuczynski: hmm, that might work.  We do return the Message to the `DoLoopTemp` API so I guess we could add an additional BSON parameter or an additional interceptor callback.

server.go
denniskuczynski: yeah, we could probably optimize this as it is really a BSON UUID.  I actually copied this function from our proxy code.

server.go
denniskuczynski: I'm unsure of why this API returns an array -- I'll have to check.  But I'll change this to >= 1 for sanity.

server.go
denniskuczynski: updated

server_test.go
denniskuczynski: updated

server.go
denniskuczynski: updated

inttests/proxy_grpc_sanity_test.go
denniskuczynski: yeah, I think the gRPC blocking should be fine here

TomerYakir: +1

denniskuczynski:
@TomerYakir I fixed the failing test (an error in our test session handler).  Ready for another look!

server.go
TomerYakir: I think it's less of a problem on our proxy code since it gets invoked on a far less hot path. That said, I assume that we won't be hitting the `reflect.DeepEqual` path of the condition too often, so this can be ignored for now

server.go
TomerYakir: Happy to resolve this - could you please file a ticket so we can follow-up?

TomerYakir (approved):

server.go
NataliaSulzhenko: this was also my concern when we did the initial review of the functionality.
@denniskuczynski can we please create a Jira to follow up on this part of the code?

denniskuczynski:
evergreen retry

server.go
denniskuczynski: 👍

server.go
denniskuczynski: 👍

PR #117, opened 2022-08-12
CLOUDP-132717 - gRPC pinned session lookup should check getMore for cursorId

denniskuczynski:
Previously we were only looking for a `cursor` sub-document when looking for pinned sessions which was actually the response structure, not the request structure.  We should have been looking for `getMore` being passed the cursorId as its argument.


server.go
```diff
@@ -591,6 +591,7 @@ func (s *GRPCServer) lookupPinnedSession(SSLServerName string, stream grpc.Serve
 		}
 
 		// check for a cursor id
+		var cursorId int64
 		cursorIdx := BSONIndexOf(cmd, "cursor")
```
denniskuczynski, line 595: I may have had this backwards as this is the return format for specifying a cursor.  We might not need this block anymore.

TomerYakir (approved):

PR #123, opened 2022-12-21
Upgrade to Go 1.19

louisaberger:


inttests/run_integration_tests_mongos_mode.sh
```diff
@@ -28,7 +28,7 @@ $MONGO_DIR/mongod --port $((MONGO_PORT+2)) --dbpath `pwd`/dbpath/6 --logpath `pw
 $MONGO_DIR/mongo --port $MONGO_PORT --eval "rs.initiate({'_id': 'proxytest2', 'protocolVersion': 1, 'members': [{_id:0, host:'$HOSTNAME:$MONGO_PORT'},{_id:1, host:'$HOSTNAME:$((MONGO_PORT+1))',priority:0},{_id:2, host:'$HOSTNAME:$((MONGO_PORT+2))',priority:0}]})"
 
 
-sleep 10 # let replica sets reach steady state
+sleep 60 # let replica sets reach steady state
```
louisaberger, line 31: Bumping this fixed some flakiness in the remote connection tests. I _think_ it's okay to just bump it because this is just waiting for the testing fixture cluster to be fully up before we attempt to do anything. What do you think? (Also, should it be a different value?)

TomerYakir: That's fine - I sent you another suggestion over Slack

TomerYakir: looking at the logs I think the proxy server might be panicking at some uncaught place. We do capture panics on DoLoopTemp but maybe it panics at a different place in the stack?

louisaberger: Note -- I confirmed that this is also flaky on Go 1.18, so it looks unrelated to the upgrade. I'll file a separate ticket to fix some of the issues there.

louisaberger:
@TomerYakir RFAL

PR #124, opened 2023-01-27
CLOUDP-155724: Add a way to stop mongonet TCP and GRPC Servers

pstefek-mongo:
Jira: https://jira.mongodb.org/browse/CLOUDP-155724
Proxy EVG: https://spruce.mongodb.com/version/63d947b99ccd4e048d4fb945/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


proxy.go
```diff
@@ -234,6 +235,17 @@ func (p *Proxy) Run() error {
 			defer wg.Done()
 			grpcErr = p.gRPCServer.Run()
 		}()
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for {
+				select {
+				case <-p.Context.Done():
+					p.gRPCServer.GracefulStop()
```
pstefek-mongo, line 244: You might wonder, what happens if `p.gRPCServer.GracefulStop()` is called before `p.gRPCServer.Run()` The gprc server package is aware of this case and will just error for fail. https://github.com/grpc/grpc-go/blob/c1d7d7acc3d14788d07ded8be9cd74c136084294/server.go#L775

proxy.go
```diff
@@ -234,6 +235,17 @@ func (p *Proxy) Run() error {
 			defer wg.Done()
 			grpcErr = p.gRPCServer.Run()
 		}()
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for {
+				select {
+				case <-p.Context.Done():
+					p.gRPCServer.grpcServer.Stop()
```
pstefek-mongo, line 244: You might wonder, what happens if `p.gRPCServer.grpcServer.Stop()` is called before `p.gRPCServer.Run()` The gprc server package is aware of this case and will just error for fail. https://github.com/grpc/grpc-go/blob/c1d7d7acc3d14788d07ded8be9cd74c136084294/server.go#L775

server.go
```diff
@@ -291,8 +291,8 @@ func (s *TCPServer) Close() {
 	<-s.doneChan
 }
 
-func NewTCPServer(config TCPServerConfig, serverWorkerFactory ServerWorkerFactory, loggerFactory LoggerFactory) TCPServer {
-	ctx, cancelCtx := context.WithCancel(context.Background())
+func NewTCPServer(config TCPServerConfig, serverWorkerFactory ServerWorkerFactory, loggerFactory LoggerFactory, ctx context.Context) TCPServer {
+	ctx, cancelCtx := context.WithCancel(ctx)
```
pstefek-mongo, line 295: For the tcpServer just allow us to pass our own cancelable parent context in

pstefek-mongo:
@denniskuczynski for the grpc
@tolsen for general multi threading and the tcp server

proxy.go
```diff
@@ -229,10 +230,23 @@ func (p *Proxy) Run() error {
 		}()
 	}
 	if p.gRPCServer != nil {
+		cancelCtx, cancelFn := context.WithCancel(p.Context)
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
 			grpcErr = p.gRPCServer.Run()
```
pstefek-mongo, line 237: There seems to be no easy way I found to pipe a context into the grpc server so I decided to create a separate goroutine which just to see if the context is cancelled and if so kills the server

proxy.go
pstefek-mongo: This is also fine to call if the server has already been shutdown

tolsen (changes requested):
Please change the PR description to describe the change, and not the problem.

Otherwise the code looks fine.  Well, I see some old code that may be a context leak, but it's not new code. lol

proxy.go
denniskuczynski: I think it would probably be cleaner from an API perspective to put the context waiting in the GrpcServer's Run method and pass the context into the constructor to be symmetric with the TCPServer impl.

pstefek-mongo: done!

server.go
```diff
@@ -526,6 +531,22 @@ func (s *GRPCServer) Run() error {
 		return err
 	}
 
+	defer s.cancelCtx()
```
denniskuczynski, line 534: I may not be familiar with the `context` API, but I don't think we want to explicitly cancel this context within here -- I think we want the users who pass a context value to `NewGRPCServer()` to be able to cancel that context.  So I don't think we need `cancelCtx` within here.  We just want to ensure that we listen to the `Done()` channel to shutdown the server. Does that make sense?

tolsen: This is a subcontext that he is cancelling.  It is created on line 342.

We have to cancel it at some point, since we are creating it inside NewGRPCServer().  Otherwise we will have a context leak.

Cancelling subcontexts do not affect the parent context.  The reverse is not true though.

pstefek-mongo: I'm with Tim here

denniskuczynski: right, but do we need to create the sub context at all?

denniskuczynski: oh, I see what you're saying.  You're using this subcontext to cleanup the new goroutine listening for the parent cancellation.  👍

denniskuczynski: and that wouldn't happen if they didn't pass in a cancellable parent context

pstefek-mongo: or they just didn't cancel it and the server stopped for some other reason (error, etc)

denniskuczynski (approved):
LGTM 👍

tolsen (approved):

PR #125, opened 2023-02-06
CLOUDP-156782 - capture mongonet test panics

TomerYakir:


inttests/multiple_proxies_test.go
```diff
@@ -47,7 +47,9 @@ func TestProxyMongosModeProxyProtocolInt(t *testing.T) {
 	if ok, _, _ := proxy.OnSSLConfig(nil); !ok {
 		panic("failed to call OnSSLConfig")
 	}
-	go proxy.Run()
+	if err := proxy.Run(); err != nil {
```
TomerYakir, line 50: `proxy.Run()` already starts a goroutine, so no need to spin another one

PR #127, opened 2023-02-23
CLOUDP-152769 : use bidirectional streams 

NataliaSulzhenko:


server_test.go
```diff
@@ -396,135 +399,32 @@ func TestGRPCServer(t *testing.T) {
...
-	if err != nil {
+	if err = sendMsg(t, stream, find); err != nil {
 		t.Error(err)
 	}
-
-	// Shutdown
-	server.GracefulStop()
-}
-
-func TestGRPCServerCursorIdNotFound(t *testing.T) {
```
NataliaSulzhenko, line 472: removed as we do not pin Cursors in mongonet anymore

server_test.go
```diff
@@ -396,135 +399,32 @@ func TestGRPCServer(t *testing.T) {
...
-	if err != nil {
+	if err = sendMsg(t, stream, insert); err != nil {
 		t.Error(err)
 	}
-
-	// Shutdown
-	server.GracefulStop()
-}
-
-func TestGRPCServerTransactionNotFound(t *testing.T) {
```
NataliaSulzhenko, line 418: removed as we do not pin transactions in mongonet anymore

denniskuczynski (commented):
First pass looks good to me.  Will take another pass with the code checked out.

inttests/proxy_grpc_sanity_test.go
```diff
@@ -39,124 +37,58 @@ func TestProxyMongosModeGRPCSanity(t *testing.T) {
...
 	conn, err := grpc.DialContext(goctx, grpcHostPort, grpc.WithInsecure(), grpc.WithBlock())
 	if err != nil {
 		t.Error(err)
 	}
 	defer conn.Close()
 
+	ctx := context.Background()
+	stream, err := openStream(ctx, conn)
+	if err != nil {
+		t.Error(err)
```
TomerYakir, line 51: [nit] you may want to use `t.Fatal` here as I don't believe there's a point to run the rest of the test code if this fails

inttests/proxy_grpc_sanity_test.go
```diff
@@ -3,17 +3,15 @@ package inttests
...
-	"time"
-
 	. "github.com/mongodb/mongonet"
 	"github.com/mongodb/mongonet/util"
 	"github.com/mongodb/slogger/v2/slogger"
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/bson/primitive"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/metadata"
+	"io"
```
TomerYakir, line 13: [trivial] dependency ordering

session.go
```diff
@@ -99,14 +103,6 @@ func (s *Session) SetPrivateEndpointId(id string) {
 	s.privateEndpointInfo.privateEndpointId = id
 }
 
-func (s *Session) IsSticky() bool {
```
TomerYakir, line 102: [q] curious - how would downstream servers know to reuse a connection for things like `getMore`?

server_test.go
```diff
@@ -627,18 +535,18 @@ type TestSessionWithContext struct {
...
+func (tsc *TestSessionWithContext) DoLoopTemp(m mongonet.Message) error {
 	atomic.AddInt32(tsc.counter, 1)
 	ctx := tsc.ctx
 
 	for {
 		m, err := tsc.session.ReadMessage()
 		if err != nil {
 			if err == io.EOF {
-				return
+				return err
```
TomerYakir, line 546: [q] is `EOF` considered an error?

inttests/proxy_grpc_sanity_test.go
```diff
@@ -209,10 +146,15 @@ func TestProxyMongosModeGRPCTransaction(t *testing.T) {
...
-	_, err = sendAndRecvMsg(t, conn, insert)
+	if err = sendMsg(t, stream, insert); err != nil {
+		t.Error(err)
+	}
+
+	resp, err := receiveMsg(t, stream)
 	if err != nil {
 		t.Error(err)
 	}
+	t.Logf("Insert responce received: %v", resp)
```
TomerYakir, line 157: [sp] responce -> response (note there's multiple occurrences of this)

server.go
```diff
@@ -547,89 +385,138 @@ func (s *GRPCServer) Run() error {
...
+func (s *GRPCServer) listen() (net.Listener, error) {
+	var lis net.Listener
+	var err error
+	if s.config.grpcUnixDomainSocketFile != "" {
+		lis, err = net.Listen("unix", s.config.grpcUnixDomainSocketFile)
+		if err != nil {
+			s.logger.Logf(slogger.ERROR, "Error starting local gRPC server over unix socket %v ", s.config.grpcUnixDomainSocketFile)
+			return nil, err
+		} else {
+			s.logger.Logf(slogger.WARN, "Starting local gRPC over unix socket %v ", s.config.grpcUnixDomainSocketFile)
```
TomerYakir, line 405: [q] extraneous trailing space?

server.go
```diff
@@ -346,191 +333,42 @@ func NewGRPCServer(config GRPCServerConfig, serverWorkerFactory ServerWorkerFact
 		loggerFactory: loggerFactory,
 		workerFactory: serverWorkerFactory,
 		grpcServer: grpc.NewServer(
-			grpc.ForceServerCodec(BytesCodec{}),
+			grpc.ForceServerCodec(MongoDBCodec{}),
+			grpc.MaxSendMsgSize(48000000),
```
TomerYakir, line 337: [nit] have 48000000 as const?

server.go
```diff
@@ -346,191 +333,42 @@ func NewGRPCServer(config GRPCServerConfig, serverWorkerFactory ServerWorkerFact
...
+			grpc.MaxRecvMsgSize(48000000),
 			grpc.KeepaliveParams(keepalive.ServerParameters{
-				Timeout: time.Minute * 5,
+				Timeout: time.Minute * 60,
```
TomerYakir, line 340: [q] why are we bumping the keepalive timeout?

server.go
```diff
@@ -547,89 +385,138 @@ func (s *GRPCServer) Run() error {
 			}
 		}
 	}()
-	s.logger.Logf(slogger.WARN, "Starting local gRPC server %v\n", lis.Addr())
+	s.logger.Logf(slogger.INFO, "Starting local gRPC server %v\n", lis.Addr())
```
TomerYakir, line 388: [q] do we need the trailing `\n` here?

server.go
```diff
@@ -547,89 +385,138 @@ func (s *GRPCServer) Run() error {
...
+	return s.handleStream(stream, false)
+
+}
+
+func (s *GRPCServer) AuthenticatedStreamHandler(srv interface{}, stream grpc.ServerStream) error {
+	return s.handleStream(stream, true)
+}
+
+func (s *GRPCServer) handleStream(stream grpc.ServerStream, isAuthenticated bool) error {
+	//TODO: add metadata check
```
TomerYakir, line 433: [q] are you planning to handle this TODO as part of this PR?

server.go
```diff
@@ -547,89 +385,138 @@ func (s *GRPCServer) Run() error {
...
+	s.logger.Logf(slogger.INFO, "*** Accepted a GRPC server stream connection isAuthenticated: %v", isAuthenticated)
```
TomerYakir, line 434: [q] stray debugging?

server.go
```diff
@@ -547,89 +385,138 @@ func (s *GRPCServer) Run() error {
...
-			}
-			s.logger.Logf(slogger.DEBUG, "Looking up session by cursorId %v/%v", SSLServerName, cursorId)
-			worker, found := s.LookupSessionByCursorId(SSLServerName, cursorId)
-			if !found {
-				return nil, m, NewMongoError(fmt.Errorf("cursor id %v not found", cursorId), CursorNotFoundErrorCode, "CursorNotFound")
-			}
-			s.logger.Logf(slogger.DEBUG, "Session found with cursorId %v/%v", SSLServerName, cursorId)
-			return worker, m, nil
+func (s *GRPCServer) processErrorAndCloseWorker(worker ServerWorker, err error) error {
+	//TODO process and make gRPC compatible Error
```
TomerYakir, line 503: [q] same question about TODO

session.go
NataliaSulzhenko: according to the [spec](https://docs.google.com/document/d/18hhq36_2m5MjtW1jShitP3zbahUFy_ePSDG8by2M454/edit#bookmark=id.v9reb4l1i81t) the downstream server should parse messages and:
```
if the command is getMore or killCursors, then use the existing stream if present
if the lsid and txnNumber exist and autocommit is false and startTransaction is not true, then use the existing stream if present
otherwise, create a new stream and cleanup the old one if present
```

server_test.go
NataliaSulzhenko: According to gRPC specification for streams (its a response from the gRPC mailing list but I can get a specification link if needed): 
```
Client-side, io.EOF is returned directly from grpc-go steram.Send() calls when the server closes the stream. 
At that point, Recv() should be called until it returns a non-nil error to receive the server's messages and eventually the status.  If io.EOF is returned by Recv(), that indicates a success; otherwise that is the final status of the RPC. 
Make sure you always call Recv() on the stream until you get a non-nil error or cancel the context used when creating the stream, or else a goroutine and other resources will be leaked.
```

inttests/proxy_grpc_sanity_test.go
NataliaSulzhenko: thanks, that is my forever typo

server.go
NataliaSulzhenko: because now we run all the commands on one stream and it takes more time that just a single call. According to the gRPC spec we should not use the timeout at all, so I probably should remove this as well

server.go
NataliaSulzhenko: creating a Jira for it, to make changes smaller, will add that in scope of error handling

session.go
TomerYakir: thanks!

server_test.go
TomerYakir: so the caller of `DoLoopTemp` is expected to check whether it's `io.EOF` and react accordingly?

server.go
NataliaSulzhenko: yes :(

server.go
denniskuczynski: Originally these keep alive settings were for the scope item:
```
Replicate current TCP keepalive settings with gRPC channel keepalive settings [gRPC keep alive specification](https://github.com/grpc/grpc/blob/master/doc/keepalive.md)
* Decided not to implement server-side, will be implemented in the Serverless Proxy instead.
```
e.g. https://www.mongodb.com/docs/manual/faq/diagnostics/#does-tcp-keepalive-time-affect-mongodb-deployments-

but it sounds like we don't need that anymore

server.go
```diff
@@ -547,89 +385,138 @@ func (s *GRPCServer) Run() error {
...
+		s.logger.Logf(slogger.WARN, "Expected RemoteAddr incoming metadata was not present")
+	}
+	session := NewGRPCSession(s, stream, serverName, remoteAddr, md, isAuthenticated)
+	worker, err := s.workerFactory.CreateWorker(session)
+	if err != nil {
+		return fmt.Errorf("error creating worker %v", err)
+	}
+	for {
+		var m []byte
+		err := stream.RecvMsg(&m)
```
denniskuczynski, line 484: Since we're no longer looking up the pinned sessions/cursors, I think we can now directly call RunWithStream and probably remove that first message optimization from the code.

proxy_session.go
```diff
@@ -115,7 +118,7 @@ func (ps *ProxySession) Stats() bson.D {
 	}
 }
 
-func (ps *ProxySession) DoLoopTemp(m Message) {
+func (ps *ProxySession) DoLoopTemp(m Message) error {
```
denniskuczynski, line 121: [q] With TCP we never returned an error here because that means we weren't able to actually send the error to the client.  I think that behavior would be the same with gRPC, so I'm not sure returning an error here has much value?

server.go
NataliaSulzhenko: Oh yes, I was considering that, but RunWithStream is executed inside a doLoop, which is surrounded by all the extra logic(common w mongoRPC) regarding metrics and message handling with underlying mongo .... so that would mean I need to reimplement it here, which I prefer to avoid

server_test.go
NataliaSulzhenko: yes, in terms of gRPC it's done in processErrorAndCloseWorker, in it's io.EOF that is means that client is done streaming and I need to close the session, otherwise, return appropriate appropriate gRPC error code based on the error

proxy_session.go
NataliaSulzhenko: in terms of gRPC if it's io.EOF that is means that client is done streaming and I need to close the session, otherwise, return appropriate appropriate gRPC error code based on the error.

```
//RunWithStream executes the doLoop which calls stream.RecvMsg(m) for all subsequent reads from the client
		//this logic is aligned with current TCP handling(including retries/and metric collection) in case of non retries errors
		//error returned here and translated to the respective gRPC error
		err = session.RunWithStream(worker, message)
		if err != nil {
			return s.processErrorAndCloseWorker(worker, err)
		}
```

server.go
NataliaSulzhenko: removing

inttests/proxy_grpc_sanity_test.go
NataliaSulzhenko: done

inttests/proxy_grpc_sanity_test.go
NataliaSulzhenko: done

server.go
NataliaSulzhenko: done

server.go
NataliaSulzhenko: removed

server.go
NataliaSulzhenko: removed

server.go
NataliaSulzhenko: removed

server.go
NataliaSulzhenko: added a Jira, in which TODO will be addressed

server.go
```diff
@@ -333,204 +322,49 @@ type GRPCServer struct {
...
 	cancelCtx     context.CancelFunc
-
-	pinnedSessionMap     map[*pinnedSessionKey]ServerWorker
-	pinnedSessionMapLock sync.RWMutex
 }
 
 func NewGRPCServer(config GRPCServerConfig, serverWorkerFactory ServerWorkerFactory, loggerFactory LoggerFactory, ctx context.Context) GRPCServer {
-	ctx, cancelCtx := context.WithCancel(ctx)
+	ctx = context.Background()
+	const m = 48000000
```
denniskuczynski, line 329: We should put this at the top of the file `const maxGrpcMessageSize = 48000000`

server.go
```diff
@@ -547,89 +381,128 @@ func (s *GRPCServer) Run() error {
...
-		cmd, err := simpleBSON.ToBSOND()
+	} else {
+		s.logger.Logf(slogger.WARN, "Expected RemoteAddr incoming metadata was not present")
+	}
+	session := NewGRPCSession(s, stream, serverName, remoteAddr, md, isAuthenticated)
+	worker, err := s.workerFactory.CreateWorker(session)
+	if err != nil {
+		return fmt.Errorf("error creating worker %v", err)
+	}
+	for {
```
denniskuczynski, line 477: I think we can remove this for loop now, since we're using the for loop within `RunWithStream`

server.go
```diff
@@ -547,89 +381,128 @@ func (s *GRPCServer) Run() error {
...
+		//RunWithStream executes the doLoop which calls stream.RecvMsg(m) for all subsequent reads from the client
+		//this logic is aligned with current TCP handling(including retries/and metric collection) in case of non retries errors
+		//error returned here and translated to the respective gRPC error
+		err = session.RunWithStream(worker)
 		if err != nil {
-			return nil, nil, err
+			return s.processErrorAndCloseWorker(worker, err)
```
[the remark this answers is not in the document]
denniskuczynski: i.e. we removed the code to reuse a worker based on data lookup

proxy_session.go
denniskuczynski: I see now 👍

server.go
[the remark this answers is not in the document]
NataliaSulzhenko: sorry not sure I understand this point, the session should be always closed by client, as per MongoDB protocol:

```
Once the server has sent its final response message, it MUST return to waiting for the client to send a message over the stream or for the stream to be terminated.
```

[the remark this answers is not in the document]
denniskuczynski: So for TCP we do:
https://github.com/mongodb/mongonet/blob/6c9dabc933270f5293a52406cd22ed2ed04ef0e8/session.go#L157
so that should be the model we're following now:

e.g. we want the ServerWorker to be closed when we won't send any more messages to it and I don't think the serverless frontend will send us more messages to the created ServerWorker we run with the stream
https://github.com/10gen/atlasproxy/blob/39d3b612f71b3831802f03f2d8ef1cb0b620f533/session_proxy.go#L244

NataliaSulzhenko:
broke tests, will investigate tomorrow

server.go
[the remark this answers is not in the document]
NataliaSulzhenko: done

NataliaSulzhenko:
> broke tests, will investigate tomorrow

tests are green

TomerYakir (approved):

PR #131, opened 2023-03-23
Add the ability to gracefully stop gRPC server

ahmedssulaiman:
Currently there is no way to trigger the grpc server to stop. We need to explicitly be able to call gracefulStop to stop the server upon shutdown. We noticed this in the UIS when shutting down the UIS still left the UDS file not being cleaned up and resulting in errors like:

```
 atlasuis panicked: listen unix /tmp/.../atlas-uis1.sock: bind: address already in use
```

This needs to be tied to the calling application's shutdown logic, see [example](https://gist.github.com/embano1/e0bf49d24f1cdd07cffad93097c04f0a#file-server-go-L73).

Anyone using this library and using the gRPC server should call this method to ensure the connection is closed and the UDS file (if present) is cleaned up.

To test this, I ran this against the UIS and restarted the application to ensure the file was cleaned up correctly and re-recreated on startup


NataliaSulzhenko (approved):
LGTM

proxy.go
```diff
@@ -241,6 +241,12 @@ func (p *Proxy) Run() error {
 	return MergeErrors(tcpErr, grpcErr)
 }
 
+func (p *Proxy) GracefulStop() {
+	if p.gRPCServer != nil {
+		p.gRPCServer.GracefulStop()
+	}
```
denniskuczynski, line 247: Should this attempt to close the TCPServer as well?

ahmedssulaiman: [Discussed offline with Dennis]

I know for TCPServer there is this method `func (s *TCPServer) Close()` but this method (even before the rename) is never being called except in tests

I can add it to the mongonet as part of close() in case the server application explicitly wants to call close on the underlying server but we need to be sure of what actually happens during existing AtlasProxy close and what happens with this change

I was also a little hesitant to touch the proxy+TCP workflow since it works. I'll file a ticket to investigate the proxy side of things.

ahmedssulaiman: Filed [CLOUDP-168285](https://jira.mongodb.org/browse/CLOUDP-168285) to investigate

PR #132, opened 2023-04-06
CLOUDP-164303: Check required gRPC Metadata and return gRPC formatted errors

NataliaSulzhenko:


server.go
```diff
@@ -424,18 +431,42 @@ func (s *GRPCServer) UnauthenticatedStreamHandler(srv interface{}, stream grpc.S
...
+			"Authentication is required for AuthenticatedCommandStream execution",
+		)
+	}
 	return s.handleStream(stream, true)
 }
 
 func (s *GRPCServer) handleStream(stream grpc.ServerStream, isAuthenticated bool) error {
-	//TODO: add metadata check in CLOUDP-164303
-	peer, ok := peer.FromContext(stream.Context())
+	md, ok := metadata.FromIncomingContext(stream.Context())
```
denniskuczynski, line 452: [q] Should we pass the metadata queried above into this method?

server.go
```diff
@@ -508,3 +546,35 @@ func (s *GRPCServer) closeWorker(worker ServerWorker) {
...
+	}
+
+	w, err := strconv.Atoi(wireVersion[0])
+	if err != nil {
+		return status.Errorf(
+			codes.FailedPrecondition,
+			fmt.Sprintf("mongodb-wireVersion has incorrct format expected int but got %v", w),
+		)
+	}
+	if w < DefaultMinWireVersion {
```
denniskuczynski, line 566: [q] Should this use `s.config.minWireVersion`?

denniskuczynski: or should we not put that in the config and just use constants?

NataliaSulzhenko: hm, I think we should use config, you are right, thank you for noticing

server.go
NataliaSulzhenko: so the metadata queried above only in case of Authenticated method execution, and it's not happening in the Unauthenticated. 

but I agree that I can improve and check for a token presence in case if isAuthenticated == true

NataliaSulzhenko: moved auth metadata check inside handleStream

server.go
NataliaSulzhenko: changed

denniskuczynski (approved):
LGTM 👍

PR #134, opened 2023-05-02
Add the ability to specify trailer metadata in Intercept methods

ahmedssulaiman:
There are no existing tests that can leverage this well so I added the integration tests in the UIS: https://github.com/10gen/atlas-uis/pull/43/files#diff-4d0f0df7deeba59979193e24c9fc43c002d154ba826b84e4448db55791eb4319


session.go
```diff
@@ -262,7 +267,7 @@ func (s *Session) RespondToCommand(clientMessage Message, doc SimpleBSON) error
 				},
 			},
 		}
-		return s.SendMessage(rm)
+		return s.SendMessage(rm, trailerMetadata)
```
ahmedssulaiman, line 270: Only setting trailer metadata for `OP_MSG`, assuming we don't need to support the other types

ahmedssulaiman:
@NataliaSulzhenko RFAL - added new metadata field changes

NataliaSulzhenko (approved):

